PR #25, opened 2015-01-03
Changed index introductory text.

ferranpujolcamins:
I've changed the introductory paragraph in the index page. I think now it's easier to translate, and more universal. In the end Mixxx is not only used for beginners :)
What do you think?

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -8,7 +8,7 @@ Mixxx |version| User Manual
 
 Mixxx is free [#f1]_ DJ software for Windows, Mac OS X and Linux. It's
 easy to use, so you will start mixing your favorite songs in no time,
-even if you've never used any DJ software before. Mixxx is also
+even if you've never DJed before. Mixxx is also
 powerful, so experienced DJs will find all the advanced features they
 would expect. You can perform with Mixxx using :term:`MIDI` and :term:`HID`
 controllers, CD players, vinyl turntables or even just with your
--- a/source/index.rst
+++ b/source/index.rst
@@ -11,7 +11,7 @@ easy to use, so you will start mixing your favorite songs in no time,
 even if you've never DJed before. Mixxx is also
 powerful, so experienced DJs will find all the advanced features they
 would expect. You can perform with Mixxx using :term:`MIDI` and :term:`HID`
-controllers, CD players, vinyl turntables or even just with your
+controllers, CD players, vinyl turntables or even just your
 computer's keyboard and mouse.
 
 Table of Contents
--- a/source/index.rst
+++ b/source/index.rst
@@ -11,7 +11,8 @@ easy to use, so you will start mixing your favorite songs in no time,
 even if you've never used any DJ software before. Mixxx is also
 powerful, so experienced DJs will find all the advanced features they
 would expect. You can perform with Mixxx using :term:`MIDI` and :term:`HID`
-controllers, or even just with your computer's keyboard and mouse.
+controllers, CD players, vinyl turntables or even just with your
+computer's keyboard and mouse.
 
 Table of Contents
 -----------------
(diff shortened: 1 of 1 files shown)


Be-ing:
I like this. :)

ferranpujolcamins:
Done :)

ferranpujolcamins:
Agree, text updated!

source/index.rst
```diff
@@ -6,10 +6,13 @@
...
-you everything you need to become a professional DJ** (except talent and hard
-work). Using Mixxx, you can rock the party with your vinyl turntables, your
-:term:`MIDI`/:term:`HID` controllers, or even just your keyboard.
+Mixxx is free [#f1]_ DJ software for Windows, Mac OS X and Linux. It's
+easy to use, so you will start mixing your favorite songs in no time,
+even if you've never DJed before. Mixxx is also
+powerful, so experienced DJs will find all the advanced features they
+would expect. You can perform with Mixxx using :term:`MIDI` and :term:`HID`
+controllers, CD players, vinyl turntables or even just your
+computer's keyboard and mouse.
```
kain88-de: Your version is already better. I would change the second and third sentence tough.

>
It's easy to use so you can start mixing your favorite songs in no time. Mixxx offers a variety of features like effects , master-sync, vinyl-scratching and many more for experienced DJs and beginners alike.


If you also link to the features that would be nice. It's more personal and says that all features can be used independent if you are a pro or not.

 I would also scratch the `computer's` in the last sentence. It is not needed to understand what you mean.

ferranpujolcamins: > It's easy to use so you can start mixing your favorite songs in no time

I'd keep the "even if you've never DJed before".


> Mixxx offers a variety of features like effects , master-sync, vinyl-scratching and many more for experienced DJs and beginners alike.

I like to enumerate and include links to specific features. However I would ditch the "beginners alike", we already "catched" them in the preceding sentence. In essence what we want to say here is that Mixxx is not a toy, thus "catching" the advanced users. How about:

> Mixxx offers all the features an experienced DJ would expect like effects, Vinyl control, Beatloop rolls, automatic Key detection and many more.

We don't have to enumerate things like master sync here, there's the "New in 1.12" section below.

Be-ing: How about:
It's easy to use so you can start mixing your favorite songs in no time, even if you've never DJed before. Mixxx offers a variety of features like automatic beat and key detection, master sync, hotcues, looping, effects and many more for experienced DJs and beginners alike.

Mentioning vinyl in this sentence is redundant with the following sentence.

kain88-de: > Mixxx offers all the features an experienced DJ would expect like effects, Vinyl control, Beatloop rolls, automatic Key detection and many more.

I changed the lines because this makes it sound like these features can only be used when you are experienced. While we usually try that everything is easily usable even for beginners. It should bring across the point that the features are suitable for beginners and experienced users.

> I'd keep the "even if you've never DJed before".

I wouldn't keep the 'even if you've never DJed before'. Just playing a track can be done by any program nothing special. The extra features like effects, loops, master-sync, multiple decks are the cool things that allow one to DJ and which differentiate mixxx from a normal music-player like banshee/clementine.

In the end it is a matter of personal taste.

ferranpujolcamins: > I'd keep the "even if you've never DJed before".

I felt we should encourage people who might think this is something too technical or complicated for them.

Keeping the sentence above ( :smiley: ), we should find a consensus in this I think:

>It's easy to use so you can start mixing your favourite songs in no time, even if you've never DJed before. Mixxx offers a variety of features for experienced DJs and beginners alike, such as automatic beat and key detection, master sync, hotcues, looping, multiple decks, effects and many more.

Be-ing: I like that, except for "multiple decks". Of course there are multiple decks, it's a DJ program. If you want to specifically mention 4 decks though, that could work there.

ferranpujolcamins:
Done.

Be-ing:
Yay, much better :)

PR #29, opened 2015-07-22
clarify use of PFL button

Be-ing:
Many beginners get confused about how to use headphones, so this explains it in explicit detail.

### DIFF
--- a/source/chapters/user_interface.rst
+++ b/source/chapters/user_interface.rst
@@ -658,6 +658,10 @@ way down.
 .. note:: Headphone cueing is only available if you have configured a
              Headphone Output in :menuselection:`Preferences --> Sound Hardware`.
 
+.. note:: If you use an external hardware mixer, plug your headphones into that
+          rather than your sound card. Use the headphone buttons on the mixer
+          rather than the buttons in Mixxx.
+
 .. _interface-eq-gain:
 
 Equalizers and Gain Knobs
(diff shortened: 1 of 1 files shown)


source/chapters/user_interface.rst
```diff
@@ -644,9 +644,13 @@ PFL/Headphone Button
    The headphone buttons of both decks in the mixer
 
 The headphone button is also known as the :term:`pre-fader listen or PFL <PFL>`
-button. Pressing this button allows you to listen and synchronize the track you
-will play next in your headphones before your audience hears the track
-(headphone cueing). You can select more than one PFL button simultaneously.
+button. Pressing this button plays the track on the headphone output at full
```
kain88-de: Can you add the sentence at the end. The main usage is the headphone-cueing which should be explained first IHMO. Useful note should come after that.

source/chapters/user_interface.rst
```diff
@@ -644,9 +644,14 @@ PFL/Headphone Button
    The headphone buttons of both decks in the mixer
 
 The headphone button is also known as the :term:`pre-fader listen or PFL <PFL>`
-button. Pressing this button allows you to listen and synchronize the track you
-will play next in your headphones before your audience hears the track
-(headphone cueing). You can select more than one PFL button simultaneously.
+button. Pressing this button allows you to listen to and synchronize the track
```
kain88-de: `to listen to the next track and synchronizing it with the currently playing track before your audience hears it.`

The `to listen to and` sounds strange to me. The only thing I can come up with to make it sound and read smoother for me is the reordering.

Be-ing:
Could someone review this?

PR #32, opened 2015-07-23
Configuration updates

Be-ing:
Update to 1.12 terminology of "audio buffer" rather than latency, correct paragraph about sample rate, and elaborate about different sound APIs on GNU/Linux.

If you're confused about the sample rate correction, read http://xiph.org/~xiphmont/demo/neil-young.html

### DIFF
--- a/source/chapters/configuration.rst
+++ b/source/chapters/configuration.rst
@@ -27,15 +27,16 @@ in- and outputs to be used.
   By default, Mixxx tries the system default first, which is most likely 44.1
   kHz. Otherwise, Mixxx will pick a different default.
 
-* **Latency**: Latency is the lag time in milliseconds that it takes for Mixxx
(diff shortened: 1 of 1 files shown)


source/chapters/configuration.rst
```diff
@@ -160,11 +162,23 @@ On Windows, if an ASIO driver is not available for your operating system, you
 can try installing `ASIO4ALL <http://asio4all.com>`_, a low-latency audio driver
 for WDM audio devices.
 
-On GNU/Linux using JACK, make sure to start your JACK daemon *before* running
+On GNU/Linux, ALSA is the simplest sound API to configure. Using ALSA will 
+prevent any other programs from using the sound card(s) that Mixxx is using. 
+JACK allows you to route audio between JACK-compatible applications in flexible
```
kain88-de: can you make this a new paragraph by entering a new line.

PR #37, opened 2015-07-24
replace "volume" with "gain" and explain how to adjust gain properly

Be-ing:
This should alleviate https://bugs.launchpad.net/mixxx/+bug/1458677


source/chapters/user_interface.rst
```diff
@@ -605,9 +605,9 @@ Headphone and Master Mix Controls
   .. note:: Don't forget to activate the :guilabel:`PFL` button on the deck you
             want to listen to in your headphones.
 
-**Head volume Knob**
-  This button adjusts your headphone volume. You can adjust the volume of a
-  single deck's signal you are listening to in the headphones with the
+**Head Gain Knob**
+  This button adjusts your headphone gain. You can adjust the gain of a
+  single deck's signal you are listening to in the headphones with the deck's
```
daschuer: We should add a hint, that this is not a hardware volume, which should be preferred if available.

source/chapters/user_interface.rst
```diff
@@ -620,14 +620,24 @@ Headphone and Master Mix Controls
   This knob allows you to adjust the :term:`balance` (left/right orientation) of
   the master output.
 
-**Volume Knob**
-  The Volume Knob controls the overall volume of the master output. Adjust
+**Gain Knob**
+  The Gain Knob controls the overall gain of the master output. Adjust
   this knob so that the :ref:`Master VU meters <interface-vu-meter>` are just
   at the peak.
```
daschuer: are just below the yellow region. 
This should offer enough headroom for mixing and applying effects. 
This matches approximately a recommended target Loudness of -16 LUFS for broadcasting.

(The yellow area it the boom region, used in movies for explosions ans such)  

Related:
https://bugs.launchpad.net/mixxx/+bug/1429760
We may also consider replace the legacy VU meters with modern loudness meters with LUFS scale.

source/chapters/user_interface.rst
```diff
@@ -620,14 +620,24 @@ Headphone and Master Mix Controls
...
 
-  .. hint:: If the Peak indicator on top of the Master VU meter flashes, the
-            master output signal is clipping (too loud). Lower the volume with
-            using the volume knob.
+  .. warning:: Gain is not the same as volume. Improperly setting the gain too
```
daschuer: as a hardware volume

source/chapters/user_interface.rst
```diff
@@ -620,14 +620,24 @@ Headphone and Master Mix Controls
...
+               high or too low will decrease your sound quality. If you want to
+               adjust your overall volume, do so by adjusting the gain as close
+               to the sound output as possible. If you can adjust a PA system's
+               amplifier, adjust it there. This is not always an option for DJs,
+               so the next best place is usually the DJ mixer. If there is no DJ
+               mixer, adjust your soundcard. If your soundcard does not have any
```
daschuer: I would recommended this: Adjust you amplifier to the maximum volume you need during the night, (all other gains at unity, and volume sliders at Max. 
Set the current volume either at a analog DJ Mixer or at the Volume panel of your OS (eg alsamixer). 
If you cannot effort this use the Master gain to reduce the volume.

source/chapters/user_interface.rst
```diff
@@ -666,15 +676,28 @@ Equalizers and Gain Knobs
...
 
 **Gain Knob**
-  Above these knobs, the gain knob allows you to adjust the gain of the deck. In
-  general, you should adjust this knob so that at full-volume the deck's audio
-  is just at the peak of the center VU meters. This is so you can achieve the
-  widest dynamic range for your track.
+  Above these knobs, the gain knob allows you to adjust the gain of the deck. 
+  Use this to compensate for the differences in recording levels between tracks.
+  In general, you should adjust this knob so that at full-volume the deck's
+  audio is just at the peak of the center VU meters. This is so you can achieve
```
daschuer: The same as above, do not use the yellow region. This is the headroom for effects, mixing and dynamic peaks of the tracks.

source/chapters/user_interface.rst
```diff
@@ -666,15 +676,28 @@ Equalizers and Gain Knobs
...
+  the widest dynamic range for your track.
+  
+  .. note:: By default, Mixxx automatically adjusts the gain of tracks to sound
+            like they are at approximately the same volume. Your tracks must be
+            :ref:`analyzed<library-analyze>` in advance to take advantage of
```
daschuer: Your tracks must be... they are automatically analyzed if ..

source/chapters/user_interface.rst
```diff
@@ -666,15 +676,28 @@ Equalizers and Gain Knobs
...
+            this feature.
 
 **EQ Knobs**
   The low, mid, and high knobs allow you to change the filters of the audio.
   This allows you to selectively reduce or boost certain frequency ranges of
   the audio.
+  
+  .. warning:: Be careful not to raise the EQs so much that any of the VU meters
+               are going into the red. At that point, your audio is clipping and
+               will sound worse. This is more often an issue with the low EQ
```
daschuer: This is more often an issue with the low EQ ... no: all EQ sliders behave the same. You see the effect better in the VU meter for low. Thats why we have a separate clipping LED that doesn't lie.

source/chapters/user_interface.rst
```diff
@@ -666,15 +676,28 @@ Equalizers and Gain Knobs
...
+               than the mid or high EQs. If you want to emphasize a frequency
+               band, it is better to lower the other ones and boost the track's
+               overall gain instead.
```
daschuer: This is the use for the yellow region. .. for momentary band emphasize. 
If you need a permanent band boost, you should lower the gain or the target loudness in normalization preferences in a way that the yellow region is not used.

source/chapters/user_interface.rst
Be-ing: Are you suggesting to switch the order of the sentences?

daschuer: Sorry, now my comment confuses me as well. Here some new thoughts: 
Use Loudness instead of volume. 
Mixxx does not adjust THE gain knob, it adjusts an additional hidden gain.
If the track already has a replaygain info in its metadata, this will be used.  
Not sure if this all matters here.  

Draft, please impove!): 

By default, Mixxx automatically applies an additional ReplayGain for an equal loudness at unity gain.  
Missing ReplayGain values are calculated during track load or in advandes by :ref:`analyzed<library-analyze>`

Be-ing: But Mixxx doesn't apply the newly calculated value upon the first load of an unanalyzed track, correct?

daschuer: It is applied if the track is (still) paused,

Be-ing:
After a lot research and real-world experimentation, I have added a new section to the "DJing With Mixxx" chapter and rewrote the Mixer section in the "Overview of the Mixxx Interface" chapter to refer to it.

source/chapters/djing_with_mixxx.rst
```diff
@@ -303,6 +303,51 @@ and click the :guilabel:`Play` button next to the waveform.
...
+example, if your sound card is turned down and you turn the gain up on your mixer to compensate, the signal-to-noise
+ratio of the sound card output will be low and the mixer's gain will amplify
+both the signal and the noise from the sound card. The end result will have more
+noise than if the output of the sound card was turned up before the signal
+reached the mixer and the mixer's gain did not have to be turned up.
+
+.. note:: To adjust the output volume of a sound system while maintaing a
+             high signal-to-noise ratio, the gain should be
+             adjusted as close to the speakers as possible. Refer to
+             the :ref:`Gain Knob <interface-gain-knob>` section for details.
```
daschuer: The tern "gain" is confusing for me here. Here an attempt to improve it: 

To adjust the output volume of a sound system while maintaing a
high signal-to-noise ratio, use the preferably an analogue control 
as close to the speakers as possible. 
Refer to the :ref:`Gain Knob <interface-gain-knob>` section for details.

source/chapters/user_interface.rst
```diff
@@ -605,29 +606,55 @@ Headphone and Master Mix Controls
...
-
-  .. hint:: If the Peak indicator on top of the Master VU meter flashes, the
-            master output signal is clipping (too loud). Lower the volume with
-            using the volume knob.
+.. _interface-gain-knob:
+  
+**Gain Knob**
+  The Gain Knob controls the gain of the signal that Mixxx sends to your
+  sound card. Generally, this knob should not be adjusted. To change the output
+  volume, a gain control should be adjusted as close to the speakers as possible
```
daschuer: an analogue control (same as above)

Be-ing:
Why does it matter if the gain is analog or digital? Don't the same principles still apply?

source/chapters/user_interface.rst
```diff
@@ -605,29 +606,55 @@ Headphone and Master Mix Controls
...
+  computers), adjust the sound card settings with your :term:`OS <operating system>`
+  mixer program. The Master Gain Knob in Mixxx should be a last resort for
+  adjusting volume.
+  
+  .. warning:: **In no case should any part of the signal chain be clipping,
+               indicated by a level meter going into its red region
+               or an LED labeled "clipping" turning on.**
+  
+  .. seealso:: For a more in depth explanation of how to set your gains, see
+               :ref:`Setting your levels properly (gain staging) <djing-gain-staging>`.
```
daschuer: I am struggling with "In depth", because an experienced user may expect more details, like how our VU meter behaves and what yellow and red region means, in terms of standard meters. 

.. and this reminds me that Mixxx has a "Toy VU Meter", which is not comparable to any standards. 
https://bugs.launchpad.net/mixxx/+bug/1429760

So for now, I would just remove "in depth" here. Once the bug is fixed, we can add "depth" details like 
you can found here: http://www.soundonsound.com/sos/sep13/articles/level-headed.htm

Be-ing: By "in depth", I mean that the section referred to explains the principles of gain staging.

source/chapters/user_interface.rst
```diff
@@ -666,15 +693,34 @@ Equalizers and Gain Knobs
...
+  In general, you should adjust this knob so that the track's :term:`level meter`
+  stays around the top of the green region with the loudest parts of the track
+  (the transients) briefly going into the yellow region. **Do NOT turn the
+  Gain Knob up so much that the level meter is in the red region. At this point
+  the track is clipping, which sounds bad and could damage equipment.**
+  
+  .. seealso:: For a more in depth explanation of how to set your gains, see
+               :ref:`Setting your levels properly (gain staging) <djing-gain-staging>`.
+  
+  .. note:: By default, Mixxx automatically applies an additional ReplayGain so
```
daschuer: I would put these notes above "seealso"

source/chapters/user_interface.rst
```diff
@@ -666,15 +693,34 @@ Equalizers and Gain Knobs
...
+  Above these knobs, the gain knob allows you to adjust the gain of the deck. 
+  Use this to compensate for the differences in recording levels between tracks.
+  In general, you should adjust this knob so that the track's :term:`level meter`
+  stays around the top of the green region with the loudest parts of the track
+  (the transients) briefly going into the yellow region. **Do NOT turn the
+  Gain Knob up so much that the level meter is in the red region. At this point
+  the track is clipping, which sounds bad and could damage equipment.**
+  
+  .. seealso:: For a more in depth explanation of how to set your gains, see
+               :ref:`Setting your levels properly (gain staging) <djing-gain-staging>`.
```
daschuer: We should also add a link to the normalisation settings

source/chapters/user_interface.rst
Be-ing: I intentionally avoided writing about specific numeric values for levels because of that bug.

source/chapters/user_interface.rst
```diff
@@ -605,29 +606,55 @@ Headphone and Master Mix Controls
...
-  at the peak.
-
-  .. hint:: If the Peak indicator on top of the Master VU meter flashes, the
-            master output signal is clipping (too loud). Lower the volume with
-            using the volume knob.
+.. _interface-gain-knob:
+  
+**Gain Knob**
+  The Gain Knob controls the gain of the signal that Mixxx sends to your
+  sound card. Generally, this knob should not be adjusted. To change the output
```
daschuer: "to your sound card, recording and broadcasting."

source/chapters/user_interface.rst
```diff
@@ -605,29 +606,55 @@ Headphone and Master Mix Controls
...
+  gain on the speakers (which may be labeled "Volume").
+  
+  As a DJ, you may not have access to the power amplifier. In that case,
+  adjust the volume by adjusting the gain on the equipment closest to the power
+  amplifier that you have access to. Typically, this is a DJ mixer. If you are
+  not going through a mixer, adjust the gain of your sound card. If your sound
+  card does not have any controls on it (for example, for sound cards built into
+  computers), adjust the sound card settings with your :term:`OS <operating system>`
+  mixer program. The Master Gain Knob in Mixxx should be a last resort for
+  adjusting volume.
```
daschuer: Be aware that changing Master Gain changes the broadcast and recording level as well.

source/chapters/user_interface.rst
daschuer: It would be nice, if we have a note about a recommended recording and broadcasting level:
"Recommendation:  "All gain Knobs at unity ReplayGain at -16 LUFS"

source/chapters/user_interface.rst
Be-ing: How about:

For an explanation of why you should set your gains this way, see :ref:`Setting your levels properly (gain staging) <djing-gain-staging>`.

source/chapters/user_interface.rst
```diff
@@ -666,15 +693,34 @@ Equalizers and Gain Knobs
...
-  widest dynamic range for your track.
+  Above these knobs, the gain knob allows you to adjust the gain of the deck. 
+  Use this to compensate for the differences in recording levels between tracks.
+  In general, you should adjust this knob so that the track's :term:`level meter`
+  stays around the top of the green region with the loudest parts of the track
+  (the transients) briefly going into the yellow region. **Do NOT turn the
+  Gain Knob up so much that the level meter is in the red region. At this point
+  the track is clipping, which sounds bad and could damage equipment.**
+  
+  .. seealso:: For a more in depth explanation of how to set your gains, see
```
daschuer: -"in depth"

source/chapters/user_interface.rst
```diff
@@ -922,9 +968,9 @@ Sample Deck Controls
   listen to the track in your headphones.
 
 **Gain Knob**
-  Allows you to adjust the volume of the track.
+  Allows you to adjust the gain of the track.
```
daschuer: signal level

daschuer:
Great! :-)
This is now really helpful. 

I have left some notes notes that may help to improve this even more.

source/chapters/user_interface.rst
Be-ing: Does this knob control the gain applied to the Deck 1-4 outputs as well as the Master output?

Be-ing: No, it does not. I don't think that is an issue, but I'll clarify the manual.

Be-ing: Come to think of it, why is this knob even there?

ywwg: The master gain knob is a good way to control the loudness of the music to the audience when there's no actual mixer.

source/chapters/user_interface.rst
daschuer: Nice.

source/chapters/user_interface.rst
Be-ing: Well, as explained in this section, it's not a good way to control the volume.

daschuer: The issue is, that all touches to that knob are recorded and broadcasted as well. 
If you control the volume on you Amp. The recorded and broadcasted level is untouched.

Be-ing: Why should the recorded and broadcasted level be adjusted by this? What advantage does that have over using the channel gain knobs?

daschuer:
> Why does it matter if the gain is analog or digital? Don't the same principles still apply?

Not sure if my theory is correct (please correct), but here it is:

For living room loudness my analogue volume rotary is at -55 dB which is a factor of 0.0017
I have no bearable noise.
If I crank up the volume towards 0 dB and lower the master gain to the same loudness, I hear heavy noise. 

The digital peak value is lowered from 32767 to 56 (16 bit to 6 bit) which is basically a bit crusher.
The quantisation signal noise ratio drops from 90 dB to 34 dB. 
Note, that this noise is caused by the digital gain knob itself.

A analogue potmeter on a DJ Mixer or on your soundcard controlled by the OS panel, does not suffer this issue, since it is able to reduce the quantisation noise together with the signal. The final output signal has still a 16 bit resolution.

Of cause both solutions are is not able to lower the noise in the amp itself or that is cached by the cable from the the sound card to the amp. (That's why pro systems uses balanced xlr cables which are immune against common-mode interference)

source/chapters/user_interface.rst
daschuer: We had a lengthy "headroom" discussion about it when reworking the normalisation preferences. 

With the current default setting and an external volume knob in place, you are fully right, master gain should not be touched. 

But it seams that there are users who prefer not to use ReplayGain and play the track at 0 dB.
Since most pop tracks are levelled to full digital scale, they will mediately suffer clipping once they start to mix or  use the EQs or whatever. These users can reduce master gain to heave a headroom for mixing.
(IMHO they are doing wrong, since the Mixxx engine does not expect such high level tracks.)

Be-ing:
> Note, that this noise is caused by the digital gain knob itself.

Is it? Or is it caused by the noise of the digital to analog converter in your sound card? Or both? Is there any way to tell?

Be-ing:
Does reducing bit depth introduce more noise or only lower the signal closer to the already present digital noise floor?

Be-ing:
Balanced signals are good, but they are useless against poor gain structure.

daschuer:
>>    Note, that this noise is caused by the digital gain knob itself.

> Is it? Or is it caused by the noise of the digital to analog converter in your sound card? Or both? Is there any way to tell?

Oh yes, it is not but it is also not directly caused by the digital to analog converter, since the noise is already there in the digital domain. The the quantisation noise is introduced when converting the floating point samples used in the Mixxx engine to the integer samples used by the DAC. The digital gain knob is just not able to reduce it like an analogue replacement after the DAC

daschuer:
> Does reducing bit depth introduce more noise or only lower the signal closer to the already present digital noise floor?

The later.

Be-ing:
In that case, I don't see why it matters whether a gain control is analog or digital when it comes to recommending which gain control to adjust. A digital gain control does not have to be something on the DJ's computer. It could be the gain of a digital mixer or a digital signal processor further down the signal chain.

daschuer:
Yes, right. Now I understand the word "analogue" can be misleading in this context.
  
It depends on the definition of digital. I am sure a digital mixer has a final gain/volume control unit that does not behave such stupid like our master gain. 
To feed a power amp there must be something like a digital controlled potentiometer in the final output stage. A pure digital (CPU only) gain knob would suffer the same issues we have with the Mixxx master gain.

Be-ing:
I'd presume that the output gains on digital mixers and signal processors would work like they do on sound cards. That is, the signal would go through a digital to analog converter then an amplifier before being sent out.

Be-ing:
Do we agree to keep the wording as "a gain control should be adjusted as close to the speakers as possible"?

source/chapters/user_interface.rst
Be-ing: I agree that use case does not make sense. Even considering that a good use case, there would never be a need for a gain knob that amplifies, only one that attenuates. Anything that can be adjusted or wired incorrectly will be, so I think the master gain knob should be removed. Anyway, the mailing list is a better place for this discussion.

source/chapters/user_interface.rst
Be-ing: I'm not sure if it's a great idea to mention that without the level meters showing LUFS or anything that can be directly compared to LUFS.

daschuer:
Maybe it is my personal problem, but for me a gain knob is a tiny rotary controlling internal pre-amps, while a volume slider is the thing the user should use in this case. 
That's why I would avoid the word gain here. But of cause "gain" is not wrong, so decide yourself.

source/chapters/user_interface.rst
daschuer: OK, than lets postpone it.

Be-ing:
I don't know if I've ever seen a standalone power amplifier or active speakers that use a volume slider. It is conventional to use a knob, which is often labeled "gain".

source/chapters/user_interface.rst
Be-ing: The normalization settings are not mentioned anywhere in the manual yet. Would you like to write a section for it?

daschuer: If I find time, I will write the EQ section first.

Be-ing:
Notes addressed and screenshots of level meters added. I also explained when more or better speakers are needed.

source/chapters/user_interface.rst
```diff
@@ -605,29 +606,59 @@ Headphone and Master Mix Controls
...
+  volume in your headphones, adjust your sound card's gain on the headphone
+  output. Most external sound cards have a gain knob specifically for the
+  headphone jack. If you use an onboard sound card for your headphones, adjust
+  the gain control in your :term:`OS <operating system>` mixer program. If your
+  sound card only has one gain knob that controls both the main output and the
+  headphone output, then adjust the :guilabel:`Head Gain Knob` in Mixxx to
+  control the volume of your headphones without affecting your main output.
+  
+  Some sound cards built into If your headphone output is not loud enough after turning your sound card all
+  the way up,
```
daschuer: Is this a leftover from other changes?

source/chapters/djing_with_mixxx.rst
```diff
@@ -303,6 +303,72 @@ and click the :guilabel:`Play` button next to the waveform.
...
+**If a level meter is in its red region, the signal is clipping and the gain
+should be turned down.** Some equipment doesn't have a level meter and only has
+an LED that turns on when the signal clips. Clipping means that the peaks of the
+waveform are flattened because the equipment has reached the maximum level that
+it can amplify the signal to. This distorts sound in an unpleasant way and can
+damage equipment by driving it with more power than it is designed to handle.
+Increasing the gain past the point where clipping begins (further into the red
+on a meter) will distort the signal more. If there is a limiter being used with
+the sound system (which is usually the case on large PA systems), clipping more
+will not make the sound any louder but will distort it more. If you want to make
```
daschuer: It depends on the type of limiter. If we have a compressor with a well designed parameter set, is will just reduce the signal level again without notable distortion. 

Due to its floating point engine, Mixxx can handle signal above top of digital scale until it finally uses a simple clamping limiter just before it passes the signal to the DAC.

Increasing the signal level above the limiter threshold does actually make the sound louder in terms of a loudness meter, since you put additional noise to the signal (see below). The energy passed to the speakers still increases, 
due to clamping when increasing gain. A sine signal with RMS of 0.7 is transfered to a square wave with RMS 1.
This is the additional energy that finally blow your speakers. 

```

Original signal 
     /\
    /  \        / 
   /    \      / 
  /      \    / 
 /        \  / 
/          \/


Clamped signal 

     __         
    /  \
   /    \      / 
  /      \    / 
 /        \__/ 


Adititonal noise 

_____  ____/\____ 
     \/    
```

daschuer: Ok, to much details. 

Can we just remove the following, since it is unrelated to Mixxx: 

"If there is a limiter being used with the sound system (which is usually the case on large PA systems), clipping more
will not make the sound any louder but will distort it more."

source/chapters/user_interface.rst
Be-ing: Yes.

source/chapters/djing_with_mixxx.rst
Be-ing: Yes, I'll remove it.

Be-ing:
Notes addressed and graphics of example waveforms added.

daschuer:
LGTM! Thank you for your efforts. 
Just give a hint once it is ready to merge.

Be-ing:
Ready for merging.

daschuer:
Oh, I have just noticed that I have no write access. 
@rryan: Can you enable it for me?

PR #64, opened 2018-02-26
Effects chapter

Be-ing:
Add a new chapter describing how to use the effects system. Information about specific effects and their parameters are left to tooltips in the GUI.


ferranpujolcamins:
The section "An Overview of the Mixxx Interface" describes the LateNight skin. The effects section has screenshots of the Deere skin. You may consider to include an explicit statement saying that you are explaining effects on Deere.

ferranpujolcamins:
You might also consider to label the controls that appear on the screenshots. This way it will be easy for users to know which knob is the metaknob for example. Also it will be easier to relate the metaknob when the effect unit is folded with the metaknob when the unit is expanded.

source/chapters/effects.rst
```diff
@@ -0,0 +1,114 @@
...
+depending on the skin.
+
+If a channel is assigned to multiple effect units, those effect units are
+chained together. This allows you to chain more than 3 effects at a time.
+
+The mix knob crossfades between the dry signal (input to the unit) and the wet
+signal (output of the last effect in the unit). This adjusts the level of all 3
+effects in the unit together. When the knob is fully left, no sound will be
+heard from the effect unit. When the knob is fully right, only the output of
+the effect unit will be heard without any of the dry input signal.
```
daschuer: Note: The Mix knob only works for effects with not group delay. If it is uses on effects with a group delay like filter or EQs it will introduce a kind of phaser effect.

source/chapters/effects.rst
```diff
@@ -0,0 +1,114 @@
...
+to the metaknob.
+
+For information about specific effects and their parameters, hover your mouse
+over them to show the tooltip. If you do not see tooltips, check that you have
+them enabled for the skin in :menuselection:`Preferences --> Interface`.
+
+Effects (except for Reverb) that have a length of time as a parameter are
+synchronized to the tempo of decks. When these effects are in a unit which is
+assigned to a channel that Mixxx does not know the tempo of (any channel other
+than a normal deck), the :ref:`master-sync` tempo is used.
```
daschuer: Is this actually working?

Be-ing: No, this has not been implemented yet. I can remove it for now.

Be-ing:
Thanks everyone for the review. I'll finish this up soon.

@kazakore:
> Is it worth mentioning here that further routing options are available?

In the Effects chapter, yes. In the overview chapter, I don't think so.

I don't really like the language of "adjusting the balance" for the mix knob because that could be confused with adjusting the stereo balance.

source/chapters/effects.rst
kazakore: I did actually wonder about this the other day. Obviously any effect takes time to do the processing, this would usually be compensated for within a DAW (Plugin Delay Compensation , or PDC) but things can get a little complicated when effects may be turned on/off and when the live performance aspects means additional delay is best kept to a minimum. How did you decide to dea with this? Use no PDC but make sure to use effects that introduce minimal processing delay? Have a small but fixed PDC amount to hopefully cover most outcomes without overly affecting timing at output. From the above comment it obviously isn't calculated for the actual effects path in use and i would guess you have gone for the minimal delay possible due to the importance of this with live performance...
It may be worth something revisting and thinking about for v2.2 anyway. I didn't know which effects would exhibit the symptoms the worst but the current system seemed reasonable with a very quick test with some random effects in the chain anyway.

kazakore:
@Be-ing  I can understand that. I think then that "adjusting the mix of" would be better than the "crossfades between" of the original text. It also emphasises why it's called the Mix knob.

source/chapters/effects.rst
Be-ing: We cannot do delay compensation because effect units have inputs that are switchable during performance. Changing which channels have a delay applied during a performance would cause an audible gap in the signal which would be heard as a pop.

kazakore: Actually it's the delay compensation that prevents the pop/click as it is applied all the time when the effect is in the chain whether the effect is currently active or not (within a DAW anyway.) Having no compensation and turning on and off effects then there is noting which takes into account that the new signal path, with the now enable effect, takes longer to process. The question is whether the delay amount is fixed and always constant, or if it's based on effects loads into some or all FX chains (as theoretically you could have all four chains affecting one deck) or if a fixed value to hopefully cover most bases, but not actually calculated, may be used.

There are also ways in which delay time can be reduced, such as a certain amount of look ahead processing (which assumes no parameters changed) and a minor delay (1-2ms max) between changing a parameter and the difference actually reaching the audible output used in conjunction with the look-ahead processing.

daschuer: We have two deal with two types of delays. The time that it takes to process the effects.
This is included in the Mixxx audio buffer of default 20 ms.
In this the samples are fetched, processed and pushed to the soundcard.
If there is enough time to process also the samples inside all chained effects everything works nice.
A Daw application with external effects deal with it, as described above.

The other delay effect us the group delay of an effect. 
Unfortunately in most iif filters, this is not constant for all frequencies.
This is also called phase shift.
For a working Mixxx knob, the dry signal has to be processed with a allpass filter with the same phase shift. This is very advanced and currently not implemented.
Is the phase shift 180° the frequency is killed producing an unwanted phase effect.
Yo can try it yourself if you use the mix knob along with a linked reily effect in Mixxx.

kazakore: >The time that it takes to process the effects.
This is included in the Mixxx audio buffer of default 20 ms.

This is the main question I was referring to, sorry about the confusion. 20ms _sounds_  a lot to me, I know in the past I start noticing lag on playing live into music software with a buffer of around 8ms (which admittedly is one way, so more like 16ms return delay.) 6ms is usually the magic number given for when musicians will start to notice and complain. It's the rough delay of hitting a piano key and hearing the sound or the equivalent of standing roughly six feet away from your guitar cabinet. I guess this is why so many people use the Quantise mode. Something I don't really like doing on the whole though....

>Mixxx audio buffer of default 20 ms

Out of interest is it actually possible to change this anywhere? I have often wondered what the 21.3ms Audio Buffer reported near the top at the Audio Preferences page referred to, when at the bottom it says System Reported Latency of 10.6667ms and Jack currently says 16ms (as I'm not running my external audio device, which is set to 8ms according to Jack. I also wonder if the discrepancy here is due to using 3 periods/buffers, which I have always found more stable and reliable across devices but can't remember exactly what it means now.) The 21.3 figure never seems to change though....

>For a working Mixxx knob, the dry signal has to be processed with a allpass filter with the same phase shift. This is very advanced and currently not implemented.

Yeah FIR, IIR and allpass filters is definitely a fairly complicated subject and one which I have forgotten much of what I taught myself a few years back. Now I know it's mainly the EQs and Filter which may still have audible effects, as you have the processing delay (at least mostly) sorted, I will do some testing for my personal interest of whether I feel it may be worth keeping the frequency related effects on their own dedicated fx chains with the Mix knob always at 100% wet.

daschuer: > I have often wondered what the 21.3ms Audio Buffer reported near the top at the Audio Preferences page referred to ...

This is the chunk of samples Mixxx processes at once. If the processing takes longer than this time you get underflows. The Latency usage meter in the skins gives some hints if this is the case,

>System Reported Latency

This works quite well in case of ALSA. 
In the Jack case that seems to be broken. This is probably because Mixxx uses Jack via Portaudio which is quite hacky. If you do not need to interface with other Jack Apps. I recommend using ALSA directly.  
I have started to work on a native Jack integration, but unfortunately the work has stalled. 
See: https://bugs.launchpad.net/mixxx/+bug/1172299

Be-ing: Back to the manual, how about adding this:
"Note that equalizer and filter effects create a phase shift. If the mix knob is anywhere in the middle between dry and wet and any of these effects are enabled in the unit, there will be comb filtering. For more background information, read [Equalizers and Phase Shift](http://ethanwiner.com/EQPhase.html)."

daschuer: there will be additional unintended comb filtering.

Be-ing: We can't assume with certainty what the user intends. For all we know someone might want the comb filtering. I don't know why someone would want this, but people want Mixxx to do all sorts of things.

daschuer: there will be additional probably unintended comb filtering.

Thats the essence of the warning, that this is probably unintended and surprising.

rryan:
test this please

rryan:
test this please

rryan:
test this please

source/chapters/effects.rst
rryan: Reminder to remove.

Be-ing:
Do not merge quite yet, I will retake the screenshots.

Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
+SPHINXBUILD   = sphinx-build-3
```
esbrandt: How is this related to the PR?
This change breaks the build for me
``The 'sphinx-build-3' command was not found.``

PR #75, opened 2018-10-09
Several build fixes and updating of translations.

JosepMaJAZ:
This is the PR for the changes that have been mentioned in the #documentation stream on Zulip.

It has been divided into several commits that describe the type of changes.

Summing it up, it fixes some of the building scripts, and updates the translation texts that were still from years ago.

This is just a needed first step. Once this is merged, the english translation has to be pushed to transifex so that translators can start working on the new content.

Sideway, some ways for keeping this up-to-date need to be developed.


rryan:
add to whitelist

README.md
```diff
@@ -191,6 +191,44 @@ For more information on Translating with Sphinx, see [Sphinx i18n].
...
+apt-get purge python-pip
+apt-get install --reinstall python
+```
+Install pip manually:
+```
+curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+python get-pip.py
+```
+
+Alternatively, instead of installing pip on the system, you can also install a python-virtualenv and then use PyPI in that virtualenv.
```
rryan: I think we should probably make this part of the instructions -- I never do anything outside of a virtualenv, but I didn't want to complicate the instructions.

README.md
```diff
@@ -191,6 +191,44 @@ For more information on Translating with Sphinx, see [Sphinx i18n].
...
+
+### Windows troubleshooting
+First, in order to have "pip", you need to install python. Python 2.17.12 and onwards already include pip, but you should follow the steps to upgrade it:
```
rryan: Python 2.7.12?

make.bat
```diff
@@ -1,17 +1,14 @@
 @ECHO OFF
```
rryan: this is an auto-generated file, so I'm hesitant to edit it -- what are these changes doing/fixing? are they fixed in the latest version of sphinx? maybe we could upgrade our sphinx setup

rryan:
thank you for taking this up 🙏

make.bat
rryan: On second thought, this is just navel gazing and if it makes it work for you on Windows then let's go ahead.

README.md
JosepMaJAZ: I've never worked with a virtualenv, so I'm not the best person to write that then.

README.md
JosepMaJAZ: oups! going to fix that :D

make.bat
JosepMaJAZ: I don't know if it is autogenerated or not. But I needed to change it so that it worked with parameters. The file as it is does not expect parameters, just which type of manual to generate.

JosepMaJAZ:
Sorry for the delay answering to this. I forgot about it yesterday

rryan:
Yup, thanks!

PR #79, opened 2019-01-27
Escape backslash in macOS Application Support path

nchammas:
Without this escape, the path [currently shows up](https://mixxx.org/manual/2.2/en/chapters/advanced_topics.html#controller-wizard) without the space, which is incorrect. I guess ReST interprets the backslash as an escape sequence, even inside the code snippet.


rryan:
[Looks good](http://downloads.mixxx.org/builds/manual/prs/PR79/en/chapters/advanced_topics.html#controller-wizard) 👍

Be-ing:
Thanks for this! My friend who is working on another Sphinx project just asked me about this issue and I could give him the answer because I remembered this PR. :)

PR #89, opened 2020-02-19
Add versioning support

Holzhaus:
I based this on PR #87, but it should also work without it.

PR #98, opened 2020-02-27
PDF build changes

Holzhaus:
Follow up for #89 that improves the PDF builds.


Holzhaus:
Added back to PR #89.

PR #109, opened 2020-04-29
keyboard mappping: fix AutoDJ modifier (Shift), add vector source

ronso0:


ronso0:
there was some text/script error and I'll need to update the source SVG

ronso0:
Done.

ronso0:
until now, this is just in the manual. and tue source vector was in mixxx/master.

we could create a dialog, though, much like About, that shows the picture and a hotkey table read from the (custom) keyboard mapping currently used.

Holzhaus:
I'd actually prefer the web browser approach though. We have a very good manual and Ibut many users do not know it exists. If we have a button that opens the shortcuts section of the manual in a webbrowser, at least some users will be made a aware of the manual. 
I don't think someone looks up keyboard shortcuts in a club where internet might not be available.

Be-ing:
I think we should gradually move documentation directly into the application wherever this is practical.

Holzhaus:
> I think we should gradually move documentation directly into the application wherever this is practical.

This means we have even more places where we need to keep the documentation up-to-date.

ronso0:
Eventhough I agree we should put all user documentation into the manual I think Hotkeys are a special topic. Whenever I need to know hotkeys (on rare occasions admittedly) I need them immediately and am regularly annoyed that my browser needs to start up for that, restore the previous session, load the manual, all that for information that is not necessarily correct and also available directly in Mixxx and just needs to be printed in a nice way.

As I said, it would be cool to show the hotkeys in a compact dialog which is populated from `keyboard.cfg`. A simple table would be okay for now, but once we have the sccripting engine updated we can re-enable svg scripts and use the existing mapping sheet vector source and fill in the key sequences that are in currently configured. Plus general hotkeys like Ctrl + R for recording etc.

ronso0:
https://bugs.launchpad.net/mixxx/+bug/1876108

Holzhaus:
Alright, if the vector image is populated from the actual settings at runtime, I'm all for it.

PR #105, opened 2020-04-18
Update & clean up cue mode documentation

xerus2000:
`While playing: Tapping the Play/Pause button pauses the track at the current position.`

_What a surprise_

Anyways, inspired by this intriguing revelation, I cleaned up Cue modes documentation (should Cue be capitalized within a sentence?), fixing some grammar and spelling, removing superfluous information and simplifying CUP documentation greatly. I have tested and verified the behaviors of all modes whose documentation I changed.

PR #116, opened 2020-06-18
Improve consistency in Cue Mode documentation

xerus2000:
* consistently link Cue point term
* avoid gerunds, use active consistently
* refer to Mixxx Mode from Denon Mode

Somewhat of a follow-up to #105

I was also wondering whether the Denon documentation should be shortened to omit all the parts that are equal to the Mixxx mode, but it might be too confusing to refer too much back and forth.


xerus2000:
Also, the sublist in Pioneer mode is currently broken (see https://mixxx.org/manual/latest/en/chapters/user_interface.html#using-cue-modes), this fixes that as well.

PR #119, opened 2020-07-05
Travis CI: Split up HTML/PDF builds

Holzhaus:
This now builds HTML and PDF manuals separately, as the combined build exceeds maximum build time on Travis CI. The HTML is deployed to Netlify. The PDFs are not deployed right now, but could be uploaded to downloads.mixxx.org (adding files to existing deployments are not possible on Netlify).

### DIFF
--- a/build_pdf.sh
+++ b/build_pdf.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+# Parse languages and ensure that "en" is built last
+cd source || exit 1
+LANGUAGES="$(python -c 'import conf; print(" ".join(sorted((lang for lang in conf.supported_languages.keys() if lang != "en"), reverse=True)))')"
+cd .. || exit 1
+LANGUAGES="$LANGUAGES en"
+NUM_LANGUAGES="$(printf '%s' "$LANGUAGES" | wc -w)"
+
+i=1
+for lang in $LANGUAGES
+do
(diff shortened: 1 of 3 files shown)


Holzhaus:
The alternative is to upload the PDFs somewhere, then move the HTML build to a later build stage and download them before deployment.

daschuer:
Travis does not build because of: 

Skipped as per condition: IF type != pull_request AND branch =~ /^(manual-)/

Holzhaus:
Yes, because this is a PR. For PRs we build deploy previews via netlify so I figured there's no need to pollute the build queue by building on Travis CI as well.

I pushed this to the manual-2.3.x branch of my fork and it works: https://travis-ci.org/github/Holzhaus/manual/builds/705017032

build_pdf.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# Parse languages and ensure that "en" is built last
```
Be-ing, line 2: Why does it matter that `en` is built last?

Holzhaus: Hmm, for PDF builds it actually does not matter. For website builds it does matter because we're generating the redirect rules in the same loop and en should be the fallback.

Holzhaus: I'll remove that code from the build_pdf.sh script.

Holzhaus:
CI build can be checked here: https://travis-ci.org/github/Holzhaus/manual/builds/708409413

PR #121, opened 2020-08-24
Fix warnings

Holzhaus:


Holzhaus:
Merge?

source/manual_guidelines.rst
```diff
@@ -1,272 +0,0 @@
-.. include:: /shortcuts.rstext
-
-Guidelines for Mixxx Manual writers
```
daschuer, line 3: Where else is this info?

Holzhaus: We could put it in the wiki or a Readme file

daschuer: Maybe a readme file is not accessible ..

Holzhaus: It's certainly more accessible than a file that is hidden in the source directory and isn't even part of any toctree. I think GitHub supports `rst` files, I'll just move it into the repo root and link it from the readme.

Holzhaus: Done.

daschuer (approved):
LGTM, thank you.

PR #123, opened 2020-08-27
requirements.txt: Remove unneeded fabric/pycrypto/ecdsa dependencies

Holzhaus:


rryan:
Fabric is still used for maintenance, e.g.
https://github.com/mixxxdj/manual#update-source-translations
and the fabfile.py is still there, so I think it makes sense to keep them in requirements.txt unless you switch to an alternative.

Are you trying to cut down on build time on netlify?

Holzhaus:
Yes. And also this is necessary for https://github.com/mixxxdj/mixxx/pull/3047 because pycrypto takes long to build and causes issues on some systems.

rryan:
> Yes. And also this is necessary for [mixxxdj/mixxx#3047](https://github.com/mixxxdj/mixxx/pull/3047) because pycrypto takes long to build and causes issues on some systems.

How about splitting them into a requirements-dev.txt ?

Holzhaus:
We could, but do we really need fabric? For example, `fab tx_push` is just wrapper for `tx push -s`. I think we can just remove it and update the README with the actual commands.

Holzhaus:
I updated the README. @rryan Please have a look.

Holzhaus:
Merge?

Holzhaus:
Ping

Holzhaus:
@Be-ing Merge?

PR #125, opened 2020-09-03
Build Qt Help on Travis

Holzhaus:


Be-ing:
What is needed to move forward with this? I see Netlify jobs run for this PR but nothing for Travis. Have we not enabled Travis on the manual repo yet?

Holzhaus:
Travis is disabled for PRs, because the secret vars don't exist there and we can't (and don't intend to) deploy PR builds anyway.

But before merging this i need to check how to slim down the generated Qt help file. Right now, this builds the manual for each lanaguage, 5hen compiles each language into a binary Qt help file. Because of that, each binary file contains all assets (images, etc.), so the total file size is quite big. I didn't find a way to package all languages into a single file that does not contain duplicate image files using sphinx/qhelpgenerator yet, but we should investigate this because it would bring down the file size to a fraction of the current size (probably around 10MB instead of >200MB).

PR #156, opened 2020-10-12
Netlify: Add support for deploy previews for the 2.2 branch

Holzhaus:
This will be useful, especially for reviewing all these outreachy contributions.


ywwg:
I have no way to know if this is correct but if you say it works, go for it :P. If it *doesn't* work, plz fix

Holzhaus:
> I have no way to know if this is correct but if you say it works, go for it :P. If it _doesn't_ work, plz fix

You could click the deploy preview button below this PR and check if it works :D

PR #222, opened 2020-10-15
Corrected error in Chapter 3 - Waveform Displays (2.3.x)

deborahtrez:
Under the sub section "Waveform Displays" in Chapter 3 of the manual, it says "Depending on the skin, Mixxx displays either parallel waveforms (default) or separate waveforms. Select your preferred appearance in Preferences ‣ Interface ‣ Skin." This is not right

The correct way is by clicking the gear icon in the upper right corner. Then under Skin Settings, click/toggle the Parallel Waveforms option to choose between parallel or separate waveform. 

Correction made. I've done this for 2.2.x as well. @Holzhaus , waiting for your approval...

### DIFF
--- a/source/chapters/user_interface.rst
+++ b/source/chapters/user_interface.rst
@@ -113,8 +113,9 @@ visually by aligning the beats that appear in each waveform.
    Mixxx default skin (Latenight) - Parallel waveforms
 
 Depending on the skin, Mixxx displays either parallel waveforms (default) or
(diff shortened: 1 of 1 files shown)


Holzhaus:
Thanks, no need to make a second Pull Request for the `manual-2.3.x` branch. We merge the `manual-2.2.x` branch into the `manual-2.3.x` on a regular basis, to the change will end up there eventually anyway :D

Thanks for the effor though!

PR #229, opened 2020-10-16
chapters/advanced_topics: Improve Mixxx control documentation

Holzhaus:
Improve formatting/wording and add missing `:coref:`s.

### DIFF
--- a/source/chapters/advanced_topics.rst
+++ b/source/chapters/advanced_topics.rst
@@ -229,7 +229,7 @@ This will bear some similarity to what you will find on a DJ mixer (e.g. crossfa
 
 .. mixxx:control:: [Master],balance
 
-   Adjusts the left/right channel balance on the Master output
+   Adjusts the left/right channel balance on the Master output.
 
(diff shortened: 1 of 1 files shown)


ronso0:
maybe also mass-rename `:feedback: N/A` to `:feedback: (none)` or even remove it entirely?

Holzhaus:
> maybe also mass-rename `:feedback: N/A` to `:feedback: (none)` or even remove it entirely?

I replaced it with "None". I thought about removing it, but then you can't tell if there is no feedback or if it's just not documented yet.

ronso0:
the reason I suggested was that I stumbled over `:feedback: (none)`
please rename `:feedback: (none)` also to `:feedback: None` then this is ready.

Holzhaus:
> the reason I suggested was that I stumbled over `:feedback: (none)`
> please rename `:feedback: (none)` also to `:feedback: None` then this is ready.

Done,

ronso0:
Thanks, LGTM

PR #231, opened 2020-10-17
Document new controls from saved loop feature

Holzhaus:
This adds documentation for the new controls added by PR mixxxdj/mixxx#2194.


source/chapters/advanced_topics.rst
```diff
@@ -1187,12 +1252,36 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
...
+   This control can be used for controllers that have dedicated :term:`hotcue`/saved loop pad modes.
+
+   .. versionadded:: 2.4.0
+
+
+.. mixxx:control:: [ChannelN],hotcue_X_setloop
+                   [PreviewDeckN],hotcue_X_setloop
+                   [SamplerN],hotcue_X_setloop
+
+   Identical to :mixxx:coref:`hotcue_X_setloop <[ChannelN],hotcue_X_setloop>`, but this always sets a saved loop, regardless of whether a loop is enabled or not.
```
daschuer, line 1279:
```suggestion
   Identical to :mixxx:coref:`hotcue_X_set <[ChannelN],hotcue_X_set>`, but this always sets as a hotcue of type Loop, regardless of whether a loop is enabled or not.
```

source/chapters/advanced_topics.rst
```diff
@@ -1187,12 +1252,36 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
...
+   If no loop is available, this does nothing.
```
daschuer, line 1280:
```suggestion
   If no loop is available, it sets and enables a beat loop of the current mixxx:coref:`beatloop_size <[ChannelN],beatloop_size>`. 
```

source/chapters/advanced_topics.rst
```diff
@@ -1402,11 +1491,15 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
 
    Doubles :mixxx:coref:`beatloop_size <[ChannelN],beatloop_size>`. If :mixxx:coref:`beatloop_size <[ChannelN],beatloop_size>` equals the size of the loop, the loop is resized.
 
+   If a saved loop is currently enabled, the modification is saved to the hotcue slot immediately.
```
daschuer, line 1494: Be aware, that this will destroy a saved loop if the you exit it using the loop having effect.

daschuer (changes requested):
Thank you.
the loop "type" control is missing.

source/chapters/advanced_topics.rst
Holzhaus: I don't understand you you mean. Can you rephrase?

daschuer: I was just aware of the following:

Sometimes I enable a loop in the outro, bring in the new track in and the think the loop by having it, before if fade out the last track.

This "destroyes" my original saved loop.
This is nothing new with normal loops, but for this use case saved loops are currently useless.

I wonder how competitors do it.

Holzhaus: Serato has a concept of "locked" loops that don't have this behavior.

Holzhaus:
> Thank you.
> the loop "type" control is missing. 

Done.

daschuer:
Link check fails.

Holzhaus:
> Link check fails.

Unrelated, see #285.

daschuer (approved):
Ok, than let's merge.

PR #239, opened 2020-10-22
GitHub Actions: Add linkcheck job

Holzhaus:


Holzhaus:
I used this to find the broken links in #240. If these are fixed we can merge this to make sure this doesn't happen again.

Holzhaus:
Gonna merge this now to avoid introducing broken commits again.

PR #245, opened 2020-10-23
Fixing Glossary.rst

hannydevelop:
While going through the manual for version 2.2 I noticed glossary had some unnecessary repetition. Some words occurred twice.

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -34,7 +34,6 @@ Glossary of Terms
      is useful because they allow DJs to see breakdowns far in advance.
 
    key lock
-   keylock
      With key lock enabled, the tempo of the track will change but the pitch
      remains consistent.  When unlocked, the pitch slider will speed up (or slow
      down) the track and the pitch will increase (or decrease) along with it.
@@ -337,7 +336,6 @@ Glossary of Terms
      are relieved of the task of having to compile the code by yourself.
 
    audio interface
-   audio interfaces
      An audio interface is a device that facilitates the input and output of
      audio signals to and from a computer, typically connected by USB. Almost
      all computers come with a basic audio interface built in, but a higher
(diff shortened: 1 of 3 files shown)


hannydevelop:
> While going through the manual for version 2.2 I noticed glossary had some unnecessary repetition. Some words occurred twice.

I noticed this was a small problem therefore there was no need creating an issue. Please merge pull request @Holzhaus

hannydevelop:
> > While going through the manual for version 2.2 I noticed glossary had some unnecessary repetition. Some words occurred twice.
> 
> I noticed this was a small problem therefore there was no need creating an issue. Please merge pull request @Holzhaus

Pull Request checks are failing because some files used :terms and made some typo. For Instance, where some file referenced :term:`key lock` others referenced it as :term:`keylock`. I would edit the references in those files according to what the glossary explains. For example, the glossary explains :term:`key lock`.

hannydevelop:
mixxx

hannydevelop:
I am Proceeding with fixing the files with this syntax :term:`Key lock <key lock>`

hannydevelop:
Everything has been fixed, @Be-ing and @Holzhaus please review and merge

source/glossary.rst
```diff
@@ -465,15 +462,14 @@ Glossary of Terms
...
-     :term:`operating system` to make use of the :term:`soundcards` connected to
+     :term:`operating system` to make use of the :term:`soundcards <soundcard>` connected to
      the computer.
 
    macOS
      macOS (originally named "Mac OS X" until 2012 and then "OS X" until 2016)
      is the current Mac operating system that officially succeeded the classic
      Mac OS in 2001.
 
-   EQ
```
Holzhaus, line 476: I think we should keep "EQ" *and* "equalizer" because one is an abbreviation and the other one is the full word.

Holzhaus (changes requested):
Thank you!

source/glossary.rst
hannydevelop: How about something like this? Equalizer (EQ) because if you ask me as a new user, i think the image below is an eyesore. Or while we leave EQ seperately from Equalizer, we can give an explanation like the one below.

EQ
This is the abbreviation for the :term:`equalizer`
mixxx1

Your suggestions are welcome @Holzhaus, please review

Holzhaus:
Thank you!

PR #246, opened 2020-10-23
Updating instruction on accessing the Waveforms menu item

Goodiec:
Made changes in chapter 3 to make it easier for users to access the "Waveforms" menu item. In the manual currently, it says Preferences > Interface > Waveforms, it should be Preferences > Waveforms as Waveforms isn't in the Interface settings section but on its own in the sidebar. 

I think "Options" should be added before "Preferences" (Options > Preferences) wherever such a case occur making it easier for users to find the "Preferences" menu item. If this is approved, I can add the changes to this request or create another request for it

### DIFF
--- a/source/chapters/user_interface.rst
+++ b/source/chapters/user_interface.rst
@@ -130,7 +130,7 @@ then toggling the :menuselection:`Skin Settings --> Parallel Waveforms` option.
 
 .. warning :: If you have a slower computer and notice performance issues with
               Mixxx, try lowering the frame rate or changing the waveform type
-              in :menuselection:`Preferences --> Interface --> Waveforms`.
+              in :menuselection:`Preferences --> Waveforms`.
 
 **Waveform summary**
   The big waveform summary shows the waveform envelope of the track near the
@@ -141,8 +141,7 @@ then toggling the :menuselection:`Skin Settings --> Parallel Waveforms` option.
 
   .. hint :: You can select from different types of displays for the waveform,
              which differ primarily in the level of detail shown in the
-             waveform, in :menuselection:`Preferences --> Interface -->
-             Waveforms --> Summary type`.
+             waveform, in :menuselection:`Preferences --> Waveforms --> Summary type`.
 
 **Waveform overview**
   The smaller, zoomed-out version of the waveform shows the various markers
@@ -153,7 +152,7 @@ then toggling the :menuselection:`Skin Settings --> Parallel Waveforms` option.
   Clicking somewhere on the waveform allows you to jump to an arbitrary position
   in the track. Like with the Waveform summary, you can select from different
   types of displays for the overview in
-  :menuselection:`Preferences --> Interface --> Waveforms --> Overview type`.
+  :menuselection:`Preferences --> Waveforms --> Overview type`.
 
 **Vinyl Widget**
   The line on the vinyl widget rotates if the track moves. It is similar to the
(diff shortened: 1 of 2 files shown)


ronso0:
Thanks for the fix!

> I think "Options" should be added before "Preferences" (Options > Preferences) wherever such a case occur making it easier for users to find the "Preferences" menu item. If this is approved, I can add the changes to this request or create another request for it

Hmm, actually you are right. Though, I think every user has accesssed the Preferences from the menu, so adding extra `Options -> ..` throughout the manual doesn't add much information compared to the visual clutter.
What do others think?

Goodiec:
> Thanks for the fix!
> 
> > I think "Options" should be added before "Preferences" (Options > Preferences) wherever such a case occur making it easier for users to find the "Preferences" menu item. If this is approved, I can add the changes to this request or create another request for it
> 
> Hmm, actually you are right. Though, I think every user has accesssed the Preferences from the menu, so adding extra `Options -> ..` throughout the manual doesn't add much information compared to the visual clutter.
> What do others think?

By users, I meant new users. Do you think I should ask what others think in the forum?

ronso0:
> Notice that in the hints section just after the description of Waveform summary, Summary Type is mentioned but it isn't in the Waveforms settings area.

Nice catch!
'Summary' is the old term, we now call it 'Overview'.
Would you mind checking where _Summary_ occurs and replacing it?

Goodiec:
> > Notice that in the hints section just after the description of Waveform summary, Summary Type is mentioned but it isn't in the Waveforms settings area.
> 
> Nice catch 👍
> 'Summary' is the old term, we now call it 'Overview'.
> Would you mind checking where _Summary_ occurs and replacing it?

That would mean removing the description for Waveform Summary also as there is one for Waveform Overview just after Waveform summary in chapter 3. I'll make the changes and commit shortly

ronso0:
Oh sorry, I was inattentive...
Simply changing `Preferences ‣ Waveforms ‣ Summary type` > `Preferences ‣ Waveforms ‣ Waveform type` is sufficient.

ronso0:
...and please double-check if this occurs elsewhere. Thanks!

Goodiec:
> ...and please double-check if this occurs elsewhere. Thank

> Oh sorry, I was inattentive...
> So this is plain worng:
> 
> > 'Summary' is the old term, we now call it 'Overview'.
> > Would you mind checking where Summary occurs and replacing it?
> 
> Simply changing `Preferences ‣ Waveforms ‣ Summary type` > `Preferences ‣ Waveforms ‣ Waveform type` in the hint is sufficient.

Done. Thank you

ronso0:
Thanks, I'll take a look.

A small `git` hint:
instead of `git revert [commit XYA]` right after `commit XYA` you may simply use `git reset --hard HEAD~1` to revert the first commit in order to reduce history spam.
If you committed something else in between you may look into `git rebase` which allows to squash or drop commits in the history.
Note that rebase will probably screw up existing reviews on github.

Goodiec:
> Thanks, I'll take a look.
> 
> A small `git` hint:
> instead of `git revert [commit XYA]` right after `commit XYA` you may simply use `git reset --hard HEAD~1` to revert the first commit in order to reduce history spam.
> If you committed something else in between you may look into `git rebase` which allows to squash or drop commits in the history.
> Note that rebase will probably screw up existing reviews on github.

Noted. Thank you!

ronso0:
Thank you, LGTM!

PR #254, opened 2020-10-26
Update README.md

toszlanyi:
After the experience with my commits always showed errors after pushing I added 2 more advises to the manual editing using git. Hope that helps other people and you developers to get better quality PRs. Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,8 @@ Next, install the dependencies using [pip]. From within the repository root
 
 If you do not wish to use pip:
 * [Install Sphinx], the documentation generator.
-* [Install Graphviz], graph visualization software (used to draw some diagrams)
+
+[Install Graphviz], graph visualization software (used to draw some diagrams)
 
 To upgrade dependencies using pip:
 
@@ -51,9 +52,10 @@ conventions.
 
 * Clone the repository `git clone https://github.com/mixxxdj/manual.git`
 * Install requirements `pip install -r requirements.txt` (see above)
-* [Install pre-commit](https://pre-commit.com/#install), then run `pre-commit install` to enable automatic commit checks.
+* Install [Graphviz] as requirement to build the manual locally
+* [Install pre-commit](https://pre-commit.com/#install), then run `pre-commit install` to enable automatic commit checks
 * Perform changes
-* Check changes `pre-commit run --files /path/to/your/file`
+* Build the HTML manual locally using `sphinx-build -W -b html source build/html` and check the output in the `build/html` directory in your browser
 * Commit changes `git commit -m "Insert short summary of your changes here"`
 * Push changes `git push`
 * Submit a [pull request]
--- a/README.md
+++ b/README.md
@@ -50,7 +50,10 @@ conventions.
 ## Editing the manual using git (recommended)
 
 * Clone the repository `git clone https://github.com/mixxxdj/manual.git`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,7 +50,10 @@ conventions.
 ## Editing the manual using git (recommended)
 
 * Clone the repository `git clone https://github.com/mixxxdj/manual.git`
+* Install requirements `pip install -r requirements.txt` (see above)
+* Install pre-commit `pre-commit install`
```
Holzhaus, line 54:
```suggestion
* [Install pre-commit](https://pre-commit.com/#install), then run `pre-commit install` to enable automatic commit checks.
```

README.md
```diff
@@ -50,7 +50,10 @@ conventions.
...
 * Perform changes
+* Check changes `pre-commit run --files /path/to/your/file`
```
Holzhaus, line 56: This is unneeded, it will automatically check all changed files on `git commit`.

Holzhaus (changes requested):

README.md
toszlanyi: yes it is needed because the pre-commit will only complain sth is wrong when you push a commit but will not fix or prevent pushing. so you gonna get commits that have whatever issues still in there, checks go wrong here on Git. Instead the user should only commit things if locally the pre-commit is passed. Another positive effect: local executed pre-commit will modify the file (eg deleting whitespaces).

Holzhaus: > yes it is needed because the pre-commit will only complain sth is wrong when you push a commit but will not fix or prevent pushing.

No. When you run `git commit` it checks all files in that commit automatically and rejects the commit when something is wrong. Running `pre-commit run --files`/`pre-commit run --all-files` is only necessary if you start to use pre-commit *after* you already committed/pushed changes with issues.

Btw, If you also want to prevent pushes, you can run `pre-commit install -t pre-push`.

Holzhaus: Either remove this line or replace it with:

```suggestion
* Build the HTML manual locally using `sphinx-build -W -b html source build/html` and check the output in the `build/html` directory in your browser
```

Holzhaus (changes requested):

README.md
toszlanyi: ... that is also good. Why not just `make html` ?

I got a warning using `sphinx-build -W -b html source build/html`
```
Warning, treated as error:
dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting
```

Holzhaus: > ... that is also good. Why not just `make html` ?

`make html` requires make to be installed. I don't think that's the case on most windows systems. It also doesn't specify `-W` (to break on warnings) and IMHO is just useless indirection. :smile: 

>     Warning, treated as error:
>     dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting

You need to install graphviz (e.g. via your package manager or by [downloading it on this website](https://www.graphviz.org/download/) if youre on windows).

Our README is wrong in that regard: Pip only installs the python bindings, it doesn't replace the actual graphviz installation. Can you correct that too while you're at it?

ronso0:
@Holzhaus Is this ready for merge?

Holzhaus (approved):
Yes. @toszlanyi Thank you!

PR #260, opened 2020-10-28
Netlify: Add 404 page and fix redirects

Holzhaus:
I think this might work, but we won't know until we merge :/

Related Zulip Discussion: https://mixxx.zulipchat.com/#narrow/stream/109176-documentation/topic/2.2E3.20mixxxcontrols.20chapter.20broken/near/214827616


ronso0:
thanks for the quick response!

Let's see!

PR #264, opened 2020-10-29
Native Instruments Traktor Kontrol S2 MK2: Fix markup

Holzhaus:
I noticed that this was broken.

### DIFF
--- a/source/hardware/controllers/native_instruments_traktor_kontrol_s2_mk2.rst
+++ b/source/hardware/controllers/native_instruments_traktor_kontrol_s2_mk2.rst
@@ -144,7 +144,6 @@ Deck Columns
     With :hwlabel:`SHIFT`, controls gain.
     Press to reset the Quick Effect superknob.
     Press with :hwlabel:`SHIFT` to reset gain.
-    Press and turn to change the effect chain preset loaded to the Quick Effect chain (new in Mixxx 2.4).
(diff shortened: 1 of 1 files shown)


ronso0:
LGTM. But wh is the table header so unnecessarily tall?
image

ronso0:
the `control` style is inconsistent:
image

ronso0:
> LGTM. But wh is the table header so unnecessarily tall?
> image

Aaah, the reason is that header strings are inserted as `p`:
image

Holzhaus:
> > LGTM. But wh is the table header so unnecessarily tall?
> > image
> 
> Aaah, the reason is that header strings are inserted as `p`:
> image

I think that is indended by the theme and helps highlighting the table.

ronso0:
but it doesn't add value. Headers are bold and thus distinguishable already.
But no real issue right now. I postpone that to when we polish the styles in general.

Holzhaus:
> the `control` style is inconsistent:
> image

I don't want to hardcode the font family right now. I think it's better to avoid using `:hwlabel` in headlines anyway. Fixed in https://github.com/mixxxdj/manual/pull/264/commits/6c28993456844189a31fb38a7b8cb2d51bd5d2b1.

ronso0:
Yup, that what I meant.

LGTM

PR #267, opened 2020-10-29
Add support for "main" branch

Holzhaus:


daschuer:
What is the point of maintaining Travis here?

Holzhaus:
> What is the point of maintaining Travis here? 

The versionend manual needs to be built *from the latest branch* because it uses the `conf.py` from that branch to build the other branches, too.

Before, I had quirky code to determine the newest branch by name, now we can just use `main`.

daschuer:
Ok than lets merge.

PR #130, opened 2020-09-23
update MC7000 mapping info

toszlanyi:
This is the updated mapping information correlated to
https://github.com/mixxxdj/mixxx/pull/2949

PR #268, opened 2020-10-30
sphinx-mixxx: Fix usage of ampersand in hwlabel role

Holzhaus:


ronso0:
I'm a python noob..
Where is this used so i can visually compare before/after?
I went through those 4 controllers
image

Holzhaus:
Use an ampersand (&) inside a hwlabel.

Before, the ampersand was interpreted as mnemonic (like in guilabel), and to get an actual ampersand character to show up you had to escapte it by using && instead of &.

This just removes that behaviour by subclassing `SphinxRole` directly instead of `GUILabel`.

Holzhaus:
This issue was discovered here: https://github.com/mixxxdj/manual/pull/130#discussion_r512482522

ronso0:
the & still needs to be surrounded by whitespaces for this to work:
image

Holzhaus:
> the & still needs to be surrounded by whitespaces for this to work:
> image
> 

Did you upgrade the `sphinx-mixxx` package? Run `pip install -r requirements.txt`.

ronso0:
ah okay, `**pip3** install -r requirements.txt` and now it looks as intended.
Thanks!

PR #280, opened 2020-11-04
Netlify: Allow unsafe-inline scripts to accomodate RTD Theme

Holzhaus:
Unfortunately, the search function sometimes doesn't work because the
inline script hash seems to change. Looks like we need to allow
unsafe-inline scripts until this issues has been fixed upstream:

https://github.com/readthedocs/sphinx_rtd_theme/issues/817


Holzhaus:
I'll merge this myself to fix the broken search now.

PR #281, opened 2020-11-04
requirements.txt: Bump RTD theme version

Holzhaus:


Holzhaus:
Merge? This will probably speed up build times a bit, as the theme is not tagged as safe for parallel reading in earlier versions.

Issue #270, opened 2020-10-30
add description for Vinyl and Passthrough buttons


ronso0:
...in https://www.mixxx.org/manual/latest/en/chapters/user_interface.html#vinyl-control-mode-and-cueing-controls

(after #154 is merged)

Goodiec:
@ronso0 would like to work on this, can this issue be assigned to me?

Holzhaus:
OK cool, go ahead :D

deborahtrez:
Hi @ronso0 @Holzhaus, how're you?
Is this issue meant to stay open? I saw that the PR for this was merged. Does it still need some work?

ronso0:
Hej @deborahtrez , thanks for the reminder. Closed

PR #294, opened 2020-11-10
added description for vinyl and passthrough buttons

Goodiec:
This pull request is for this issue - https://github.com/mixxxdj/manual/issues/270
Kindly review.


source/chapters/user_interface.rst
```diff
@@ -458,6 +458,18 @@ control relative mode:
   to that cue point.
 * **Hot**: The track will seek to the nearest previous :term:`hotcue` point.
 
+**Vinyl button**
```
ronso0, line 461: LGTM
Vinyl & Pass appear first in the image so I suggest to move those sections up above `Abs/Rel/Const button` so they're also described first.

ronso0 (changes requested):

source/chapters/user_interface.rst
Goodiec: Done.

source/chapters/user_interface.rst
```diff
@@ -440,6 +440,18 @@ Mixxx and is **hidden** in the default
...
+
+This button enables :term:`vinyl control`. When enabled in Mixxx, the track
+on the Mixxx deck is controlled by the external vinyl control.
+
+**Pass button**
+
+This button enables passthrough. When enabled, audio from an external vinyl
+can be played on the Mixxx deck. To be able to enable this button, you have
+to select an input device for the passthrough control in
+:menuselection:`Preferences --> Sound Hardware`.
```
Holzhaus, line 453: Isn't that true for all buttons in the section? Maybe we should move that information to to parent section or remove it completely (because it already contains a reference to the vinyl control documentation).

Holzhaus (commented):
Thanks!

source/chapters/user_interface.rst
ronso0: That is ncessary for VINYL and PASSTHROUGH only.
True, we should put move that note here and instead move the 'See also' block to the top of this section and link to https://manual.mixxx.org/2.2/en/chapters/vinyl_control.html#vinyl-control

Goodiec: Yes, the note is the vinyl and paasthrough buttons but I only mentioned it for the passthrough button, I'll correct it to include the both of them. @ronso0 don't quite get where you want the "See also" block moved to.

ronso0: `To be able to enable this button, you have
to select an input device for the passthrough control`
I agree with @Holzhaus, that info would be redundant here:
> Maybe we should move that information to to parent section or remove it completely (because it already contains a reference to the vinyl control documentation


__
> don't quite get where you want the "See also" block moved to

Right above **Vinyl button** IMO:
image

Goodiec: Done, thank you.

ronso0:
Thanks, LGTM!

Holzhaus:
> Thanks, LGTM!

Let's Squash-and-merge this, as the other commit messages are not really helpful.

ronso0:
Of course, I was checking if the original author is kept when squashing.
Done.

PR #302, opened 2020-11-12
Update hardware.rst

aishagurung:
Updated the Mixxx DJ  Hardware guide link to go directly to the desired link without a redirect


ronso0:
Thanks for those fixes.
After changes to the 2.2 branch it is merged into 2.3, so there's no need to open a PR for every branch, it rather may cause merge conflicts if the wording is not exactly the same.
That said, #303 is sufficient.

PR #307, opened 2020-11-16
chapters/controlling_mixxx: Reorder controller preset instructions

Holzhaus:
The controller preset setup was improved in Mixxx 2.3. Now, when no
preset is selected, the "Enabled" checkbox is greyed out. Hence it it's
necessary to select a mapping *before* ticking the checkbox.


ronso0:
For me (testing in dev mode with `MIDI Through Port-0`) the checkbox is always ticked as soon as a preset is selected, regardless of its previous state or if I had a preset selected before that or not.
So we can reword to "Make sure the _Enabled_ checkbox is ticked in order to use the preset.", right?

ronso0:
Thanks, LGTM

PR #312, opened 2020-12-01
appendix: fix ref to library/file-format-compatibility

ronso0:


source/chapters/library.rst
```diff
@@ -888,7 +888,7 @@ hotcues and the beatgrids), and removes links to playlists or crates.
 .. hint:: If you later decide to add some of the deleted tracks back, import
           them to the Mixxx library again, see :ref:`configuration-import`.
 
-.. _getting-started-file-format-compatibility:
```
ronso0, line 891: no prob to rename that, it's only used in Appendix

Holzhaus (approved):

Issue #272, opened 2020-10-31
update Command Line Options with explicit commands for Windows & macOS


ronso0:
currently https://mixxx.org/manual/latest/en/chapters/appendix.html#command-line-options assumes Linux.

Goodiec:
@ronso0 will like to work on this, is there a resource available to get the command line alternatives for the other operating systems?

Holzhaus:
@ronso0 Can you elaborate? Where does it asumme Linux?

ronso0:
Sorry for being unclear.
A few days ago when I was trying to help a user on the forum and also when doing remote macOS debugging on Zulip, I noticed that we have no documentation about the exact 'mixxx' base command users need to enter into the terminal prompt to which they'd append the command line options.

It would be nice to have specific examples for macOS, Linux & Windows, like
`%PROGRAMFILES%\Mixxx\mixxx.exe` / `%PROGRAMFILES(x86)%\Mixxx\mixxx.exe`  and
`/Applications/Mixxx.app/Contents/MacOS/mixxx` / `% open -a Mixxx.app`
(I don't know which is more generic = always works)

deborahtrez:
> Sorry for being unclear.
> A few days ago when I was trying to help a user on the forum and also when doing remote macOS debugging on Zulip, I noticed that we have no documentation about the exact 'mixxx' base command users need to enter into the terminal prompt to which they'd append the command line options.
> 
> It would be nice to have specific examples for macOS, Linux & Windows, like
> `%PROGRAMFILES%\Mixxx\mixxx.exe` / `%PROGRAMFILES(x86)%\Mixxx\mixxx.exe` and
> `/Applications/Mixxx.app/Contents/MacOS/mixxx` / `% open -a Mixxx.app`
> (I don't know which is more generic = always works)

Hi @ronso0 @Holzhaus , this is how I was able to open Mixxx in developer mode. I use Windows.
![image](https://user-images.githubusercontent.com/28389950/100709893-ecd3b680-33bf-11eb-858f-e970c47c5bc0.png)

but this is not mentioned anywhere in the documentation. I got this from one of the forums. 
I can add this to that chapter if you give me the go-ahead (for Windows at least). For instance the full command for  `--controllerDebug` would be `mixxx --controllerDebug`

ronso0:
Sure, go ahead! I'm sure you can collect correct info for macOS, too, if you ask on Zulip.
[...]

PR #313, opened 2020-12-02
update Command Line Options with explicit commands for Windows & macOS

deborahtrez:
In response to this (issue)[https://github.com/mixxxdj/manual/issues/272] I will be adding the base command for each platform (MacOS, Windows and Linux) onto the which the command-line options can be appended. 
@ywwg 

Added for Windows. Now obtaining for MacOS.


source/chapters/appendix.rst
```diff
@@ -296,6 +296,12 @@ file(s) on the command line. These will be loaded at start-up.  Each file you
 specify will be loaded into the next virtual deck. For a list of supported file
 types, go to :ref:`configuration-import`.
 
+To launch Mixxx using the command line:
```
Holzhaus, line 299: I think we should name the whole section "Launching Mixxx from the Command Line", then make "Command Line Options" a subsection of it.

source/chapters/appendix.rst
```diff
@@ -296,6 +296,12 @@ file(s) on the command line. These will be loaded at start-up.  Each file you
...
+For Windows OS, open the folder where Mixxx is installed and copy the file path. You can also obtain the file path by right clicking on the Mixxx icon and clicking 'Properties'. The path will usually look like ``C:\Program Files\Mixxx`` or ``C:\Program Files (x86)\Mixxx``. Change directory to this path in the command line by typing ``cd C:\Program Files\Mixxx``. Once the command prompt is at this location, type ``mixxx`` and press Enter.
```
Holzhaus, line 300: It's not necessary to `cd` to it, you can can also just type `C:\Program Files\Mixxx\mixxx --developer` directly.

Holzhaus: Please also add information about the other OSes. On Linux, you can just type `mixxx` (assuming the `mixxx` executable is in `$PATH`). @ronso0 explained in #272 how to run it on macOS. I think we should use a layout like here for the different OSes: https://manual.mixxx.org/2.2/en/chapters/getting_started.html#opening-mixxx

Holzhaus (commented):
Thank you!

source/chapters/appendix.rst
deborahtrez: Ohh yes! This looks better. Let me make the required changes.

deborahtrez: > It's not necessary to `cd` to it, you can can also just type `C:\Program Files\Mixxx\mixxx --developer` directly.

Hi @Holzhaus I tried this option and it did not seem to work for me. I tried different options but this is what i got

image

I was only successful with `cd` then `mixxx`.
What do you think?

Be-ing: I think you need quotes around the full path name so the spaces are not interpreted as separating command arguments:
```
"C:\Program Files\Mixxx\mixxx --developer"
```

deborahtrez: Also, the command specified [here](https://mixxx.org/manual/latest/en/chapters/appendix.html#developer-tools) for starting Mixxx in Developer mode from a custom resource directory with MIDI and HID logging enabled might also be OS specific.
Should I change this to state that the user can simply append `--controllerDebug --developer --resourcePath res` to the base command depending on their operating system?

deborahtrez: Hey @ronso0 I would love to get your thoughts on this as well.

Holzhaus: > I think you need quotes around the full path name so the spaces are not interpreted as separating command arguments:
> 
> ```
> "C:\Program Files\Mixxx\mixxx --developer"
> ```

That should be

    "C:\Program Files\Mixxx\mixxx" --developer

deborahtrez: > I think you need quotes around the full path name so the spaces are not interpreted as separating command arguments:
> 
> ```
> "C:\Program Files\Mixxx\mixxx --developer"
> ```

Hi @Be-ing, I have just tried that as well. It still didn't work.

image

Be-ing: The path for `cd` should not include the `mixxx` executable:
```
cd "C:\Program Files\Mixxx"
mixxx --developer
```

```
"C:\Program Files\Mixxx\mixxx" --developer
```

I think those would both work.

deborahtrez: Yes. They worked! Thank you.

source/chapters/appendix.rst
```diff
@@ -284,22 +284,59 @@ Debugger Enabled                          :kbd:`Ctrl` + :kbd:`Shift` + :kbd:`D`
...
+
+
+**Mac OSX**
+  Similarly, for MacOSX, the command will be ::
+
+    %PROGRAMFILES%\Mixxx\mixxx.exe
+
+  or ::
+
+    %PROGRAMFILES(x86)%\Mixxx\mixxx.exe
```
Holzhaus, line 313: This is for windows, not macOS :smile:

source/chapters/appendix.rst
```diff
@@ -284,22 +284,59 @@ Debugger Enabled                          :kbd:`Ctrl` + :kbd:`Shift` + :kbd:`D`
...
+  You could also choose to first ::
+
+    cd "C:\Program Files\Mixxx"
+
+  and then enter ::
+
+    mixxx
+
+
+**Mac OSX**
```
Be-ing, line 306: Apple now stylizes it as `macOS` since 2016: https://en.wikipedia.org/wiki/MacOS_Sierra

source/chapters/appendix.rst
```diff
@@ -284,22 +284,59 @@ Debugger Enabled                          :kbd:`Ctrl` + :kbd:`Shift` + :kbd:`D`
...
+    %PROGRAMFILES(x86)%\Mixxx\mixxx.exe
+
+  depending on the path in which you have Mixxx installed.
+
+**GNU/ Linux**
+  The command is ::
+
+   mixxx
+
+  (assuming the mixxx executable is in $PATH). Press Enter.
```
Holzhaus, line 322:
```suggestion
  If Mixxx is in our system's search path (``$PATH``), you can just type ::

   mixxx
```

source/chapters/appendix.rst
```diff
@@ -284,22 +284,59 @@ Debugger Enabled                          :kbd:`Ctrl` + :kbd:`Shift` + :kbd:`D`
...
+
+  and then enter ::
+
+    mixxx
+
+
+**Mac OSX**
+  Similarly, for MacOSX, the command will be ::
+
+    %PROGRAMFILES%\Mixxx\mixxx.exe
```
Be-ing, line 309: This should be `/Applications/Mixxx.app/Contents/MacOS/mixxx`

source/chapters/appendix.rst
```diff
@@ -284,22 +284,59 @@ Debugger Enabled                          :kbd:`Ctrl` + :kbd:`Shift` + :kbd:`D`
...
+
+    "C:\Program Files\Mixxx\mixxx"
+
+  You could also choose to first ::
+
+    cd "C:\Program Files\Mixxx"
+
+  and then enter ::
+
+    mixxx
```
Holzhaus, line 303: I don't think this is needed.

Holzhaus (commented):
Thanks!

deborahtrez:
> Thanks!

You're welcome.

Be-ing:
Thanks. Sometimes we ask users to start Mixxx with a command line option and they don't know how. It will be useful to refer to this.

source/chapters/appendix.rst
```diff
@@ -284,18 +284,44 @@ Debugger Enabled                          :kbd:`Ctrl` + :kbd:`Shift` + :kbd:`D`
...
+
+  "C:\Program Files\Mixxx\mixxx" --developer
+
 In addition to these options, it is possible to specify one or more music
 file(s) on the command line. These will be loaded at start-up.  Each file you
 specify will be loaded into the next virtual deck. For a list of supported file
 types, go to :ref:`configuration-import`.
 
+
+
```
Holzhaus, line 324: Unneeded newlines

Holzhaus (commented):
Just a minor issue otherwise LGTM. Thank you.

PR #314, opened 2020-12-02
Replace Controller "Preset" with "Mapping"

Holzhaus:


Be-ing:
Thanks. The next step after https://github.com/mixxxdj/mixxx/pull/2920 will be doing this in the code too.

PR #319, opened 2020-12-04
Fix tables

Holzhaus:
Reenable wrapping in responsive tables and improve the Vinyl Control signal indicator table.

### DIFF
--- a/source/_static/css/mixxx.css
+++ b/source/_static/css/mixxx.css
@@ -179,3 +179,8 @@ a:hover {
 .btn-link:visited {
     color: #b69b59;
 }
+
+.wy-table-responsive table td,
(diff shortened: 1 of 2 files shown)


ronso0:
can you point me to an example table where the improvemen is visible?

Holzhaus:
> can you point me to an example table where the improvemen is visible?

The signal quality table for vinyl control. 

Before:
https://manual.mixxx.org/2.2/en/chapters/vinyl_control.html#troubleshooting

After: https://deploy-preview-319--mixxx-manual.netlify.app/chapters/vinyl_control.html#troubleshooting

ronso0:
okay, if that's the only one I don't see a difference here, both look good.
LGTM

PR #333, opened 2020-12-28
appendix: Update actual headings for changelog versions

Holzhaus:
This makes it possible to link to the changelog of a specific version.


ronso0:
Okido! LGTM

PR #339, opened 2021-01-05
Allow failure on linkcheck

lepieru:
I changed the linkcheck on the CI. It now continue even if it failed. I need your validation about it.

I did it because www.stantondj.com is down, so we can't build.

It's nice to have a check, but we can't control the Internet :o

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -7,6 +7,7 @@ jobs:
   linkcheck:
     name: Link Check
     runs-on: ubuntu-latest
+    continue-on-error: true
     steps:
     - name: Clone repository
       uses: actions/checkout@v2


Holzhaus:
This doesn't make the linkcheck job pass regardless of the actual availability of the URL, it just makes sure that the build job is not skipped, right? 

Or will there be a green checkmark in any case now? The check is for notifying us if we have broken links in our manual. If we see a red X, we can have a look and check if we need to remove/change a link. In this case this isn't necessary so we can just ignore it. If we don't get a red checkmark, the whole check becomes pointless.

Do you know why the job doesn't run at all on this PR?

lepieru:
No, I don't. And I don't see solution right now :/

Something I want is an `allow-failure` as it is talked about here: https://github.community/t/continue-on-error-allow-failure-ui-indication/16773/12.

I convert this PR to draft.

Holzhaus:
I think the link check is independent now. I'll close this for now.

PR #351, opened 2021-02-11
added more description to Analyze your tracks

### LINKED ISSUE #350: Provide more context to "Analyze your tracks/ library"
The manual never defines what exactly analysis does and why it is important, hence it can easily be overlooked - especially by the new user. It would be great to add why users should do it in advance and what happens if they don't.

https://manual.mixxx.org/2.2/en/chapters/library.html#analyze-prepare-your-tracks

deborahtrez:
This fixes #350 

The manual never defines what exactly analysis does and why it is important, hence it can easily be overlooked - especially by the new user. It would be great to add why users should do it in advance and what happens if they don't.

@Be-ing @Holzhaus @ywwg please, kindly review :)

### DIFF
--- a/source/chapters/getting_started.rst
+++ b/source/chapters/getting_started.rst
@@ -74,7 +74,7 @@ Analyze your library
    S.Brandt <s.brandt@mixxx.org>
 
 |ic_lib_prepare| Mixxx automatically analyzes tracks the first time you load
-them in a deck. This allows you to run :term:`beatgrid`, :term:`key`, and :term:`ReplayGain` detection on tracks in advance. It also prepares the waveform overviews, detects file corruption, and calculates the BPM values.
+them in a deck. This allows you to run :term:`beatgrid`, :term:`key`, and :term:`ReplayGain` detection on tracks in advance. It also prepares the waveform overviews, detects file corruption, and calculates the :term:`BPM` values.
 
 .. seealso:: For more information, go to :ref:`configuration-bpm-detection`.
 
--- a/source/chapters/getting_started.rst
+++ b/source/chapters/getting_started.rst
@@ -82,7 +82,7 @@ Why you should analyze your audio files
 ---------------------------------------
 
 If the tracks have never been played with Mixxx before, the library columns :guilabel:`BPM`,
-:guilabel:`Duration` and :guilabel:`Key` will be empty.
+:guilabel:`Duration` and :guilabel:`Key` will be empty or unreliable (because the values are just read from the file's tags if present and might be wrong). 
 It is recommended that you analyze your audio files before
 playing live to ensure the beatgrids are correct and these fields show. Furthermore, track
 analysis takes considerable CPU power and might cause skips in the audio ---
--- a/source/chapters/getting_started.rst
+++ b/source/chapters/getting_started.rst
@@ -74,19 +74,31 @@ Analyze your library
    S.Brandt <s.brandt@mixxx.org>
 
 |ic_lib_prepare| Mixxx automatically analyzes tracks the first time you load
-them in a deck, nevertheless it is recommended that you analyze them before
-playing live to ensure the beatgrids are correct. Furthermore, track
+them in a deck. This allows you to run :term:`beatgrid`, :term:`key`, and :term:`ReplayGain` detection on tracks in advance. It also prepares the waveform overviews, detects file corruption, and calculates the BPM values.
+
+.. seealso:: For more information, go to :ref:`configuration-bpm-detection`.
+
+Why you should analyze your audio files
+---------------------------------------
+
+If the tracks have never been played with Mixxx before, the library columns :guilabel:`BPM`,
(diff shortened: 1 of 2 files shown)


source/chapters/getting_started.rst
```diff
@@ -74,19 +74,31 @@ Analyze your library
...
+
+.. seealso:: For more information, go to :ref:`configuration-bpm-detection`.
+
+Why you should analyze your audio files
+---------------------------------------
+
+If the tracks have never been played with Mixxx before, the library columns :guilabel:`BPM`,
+:guilabel:`Duration` and :guilabel:`Key` will be empty.
+It is recommended that you analyze your audio files before
+playing live to ensure the beatgrids are correct and these fields show. Furthermore, track
```
ywwg, line 87: we could also say something like  "pregenerating beatgrids gives you the chance to doublecheck the beat markers to make sure they are correct.  Errors in beatgrids can cause difficulty when syncing tracks"

source/chapters/library.rst
```diff
@@ -835,7 +835,7 @@ view, then right-click on a session's name/date to access the different features
 * **Join with previous**: Append the selected history session to the end of the
   previous one.
 * **Export playlist**: Export a session in various file formats. This allows you
-  to use the data in other applications.
+  to use the data in other applic
```
ywwg, line 838: some text got removed here

source/chapters/getting_started.rst
```diff
@@ -74,19 +74,31 @@ Analyze your library
...
-:term:`key`, and :term:`ReplayGain` detection on tracks in advance. While
-analyzing, the progress in percentage and total queue length are shown.
+How to analyze your audio files
+-------------------------------
+Once you have :ref:`imported your music library<getting-started-import-audio-files>`, press
+:guilabel:`OK` on the Preferences window.
 
-.. seealso:: For more information, go to :ref:`configuration-bpm-detection`.
+* Go to the :guilabel:`Analyze` view on the left side panel of the library.
+* To analyze a few specific tracks, hold down the :kbd:`Ctrl` button, and select the tracks that you want to analyze. Once they've been highlighted, click the :guilabel:`Analyze` button on the top right of th library section.
```
ywwg, line 97: or the user can shift-select a range of tracks

ywwg: also, typo: "th" -> "the"

ywwg, line 97: maybe also add a bullet describing the button that selects between showing only recent tracks in the analysis pane, and all tracks.

ywwg (changes requested):
thanks for clarifying this point.  I made some small comments

source/chapters/getting_started.rst
```diff
@@ -74,19 +74,31 @@ Analyze your library
    S.Brandt <s.brandt@mixxx.org>
 
 |ic_lib_prepare| Mixxx automatically analyzes tracks the first time you load
-them in a deck, nevertheless it is recommended that you analyze them before
-playing live to ensure the beatgrids are correct. Furthermore, track
+them in a deck. This allows you to run :term:`beatgrid`, :term:`key`, and :term:`ReplayGain` detection on tracks in advance. It also prepares the waveform overviews, detects file corruption, and calculates the BPM values.
```
Holzhaus, line 77:
```suggestion
them in a deck. This allows you to run :term:`beatgrid`, :term:`key`, and :term:`ReplayGain` detection on tracks in advance. It also prepares the waveform overviews, detects file corruption, and calculates the :term:`BPM` values.
```

source/chapters/getting_started.rst
```diff
@@ -74,19 +74,31 @@ Analyze your library
...
+
+.. seealso:: For more information, go to :ref:`configuration-bpm-detection`.
+
+Why you should analyze your audio files
+---------------------------------------
+
+If the tracks have never been played with Mixxx before, the library columns :guilabel:`BPM`,
+:guilabel:`Duration` and :guilabel:`Key` will be empty.
```
Holzhaus, line 85:
```suggestion
:guilabel:`Duration` and :guilabel:`Key` will be empty or unreliable (because the values are just read from the file's tags if present and might be wrong). 
```

source/chapters/getting_started.rst
deborahtrez: Alright

ywwg (approved):
lgtm

source/chapters/library.rst
```diff
@@ -837,7 +837,6 @@ view, then right-click on a session's name/date to access the different features
 * **Export playlist**: Export a session in various file formats. This allows you
   to use the data in other applications.
 
-
```
Holzhaus, line 840: Unrelated change. I suggest to use `git add -p` to avoid this in the future.

Holzhaus (approved):
Thanks.

source/chapters/library.rst
deborahtrez: This is the command I was looking for.  Thank you Jan.

PR #356, opened 2021-02-26
NI Traktor Kontrol S2 MK3 - Notes on master gain unmapping

martinkirch:
Documenting the modification from https://github.com/mixxxdj/mixxx/pull/3658

### DIFF
--- a/source/hardware/controllers/native_instruments_traktor_kontrol_s2_mk3.rst
+++ b/source/hardware/controllers/native_instruments_traktor_kontrol_s2_mk3.rst
@@ -2,8 +2,8 @@ Native Instruments Traktor Kontrol S2 MK3
 =========================================
 
 The Kontrol S2 MK3 is a two-channel controller with an integrated sound card. It has two integrated stereo outputs (line and 1/8" / 3.5 mm TRS), headphone outputs (1/8" / 3.5 mm TRS) and microphone
-inputs (1/4" / 6.3 mm TRS). The MK3 uses the standard HID protocol to send and receive signals from a computer, so it can work with Mixxx. The Kontrol S2 MK3 can run from USB bus power, but using the
-separate power supply increases the brightness of the LEDs, which is helpful for using it in daylight, and increases the volume of the headphone output.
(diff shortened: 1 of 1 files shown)


Be-ing:
Thanks!

The formatting of the table on that page could use some cleanup. If you want to make another PR for that, that would be nice.

source/hardware/controllers/native_instruments_traktor_kontrol_s2_mk3.rst
```diff
@@ -18,6 +18,10 @@ Mixxx Sound Hardware Preferences
 Mixxx mapping
 ~~~~~~~~~~~~~
 
+Since v2.3 the Master knob is intentionally not linked to Mixxx's Main Gain,
+`following NI's suggestions <https://support.native-instruments.com/hc/en-us/community/posts/360005413857-The-Traktor-S2MK3-Low-output-issue-en-us-?page=1#community_comment_360000742537>`__.
+Indeed this knob also sets the hardware gain, and mosts sets require a high setting.
```
Holzhaus, line 23: This should be a short `... versionchanged` note at the top instead.

Holzhaus: If I understand correctly, the knob shouldn't have been mapped in the first placed because it controls the hardware volume. So I think this can be made less wordy.

source/hardware/controllers/native_instruments_traktor_kontrol_s2_mk3.rst
```diff
@@ -93,7 +97,7 @@ Mixxx mapping
 | (19) Quantize    | Toggles quantization for both decks                                                                                         |                                                     |
 | button           |                                                                                                                             |                                                     |
 +------------------+-----------------------------------------------------------------------------------------------------------------------------+-----------------------------------------------------+
-| (20) Gain knob   | Adjusts the gain for the master output                                                                                      |                                                     |
+| (20) Gain knob   | Adjusts the hardware gain - it is unlinked to the software main gain since v2.3                                             |                                                     |
```
Holzhaus, line 100: This doesn't belong into the table. This is the manual for Mixxx 2.3 so there's no reason to document behavior for earlier mixxx versions.

Holzhaus:
Oops, @Be-ing already merged this. @martinkirch could you make a follow up PR?

PR #370, opened 2021-04-12
update 2.3 keyboard mapping sheet

ronso0:
Mixxx-220-Keyboard-Mapping


Holzhaus:
Why is it a PNG if we have a vector version of it?

ronso0:
I was not aware of the SVG in the manual. I'll look for an efficient workflow to create a cropped version of https://github.com/mixxxdj/mixxx/blob/2.3/res/images/templates/ic_template_keyboard_mapping_sheet.svg for the manual.
I assume we can't embed PDFs? would probably be the easiest way to just grab the PDF from /mixxxdj/mixxx/2.3/res/

Why is the SVG in `/source/_templates` while all the other (raster) images are in `/source/_static`?

And why does the manual show a SVG and "Download the image" links to a PNG?
Is it to be expected that there are user viewing the manual while not having an SVG viewer? Shouldn't we offer a PDF instead?
(hmm, I recently noticed that by default there is no PDF viewer installed, even the web browser Edge can't handle them, so maybe that makes sense)

Holzhaus:
> I assume we can't embed PDFs?

No. 

> Is it to be expected that there are user viewing the manual while not having an SVG viewer? Shouldn't we offer a PDF instead?

There are two versions of the manual:
- The Web version. You need a web browser for that, which can also displays SVGs.
- The PDF version. The SVG is transformed into native PDF paths when generating the PDF.

I think that suffices.

ronso0:
okay, thanks for the clarification.
I update the SVG, too, as well as the source in https://github.com/mixxxdj/mixxx/pull/3785

I also added a note to Readme.md that the source SVG mixxxdj/mixxx should be edited first.

ronso0:
both the SVG and the downloadable PNG are updated according to https://github.com/mixxxdj/mixxx/pull/3785

PR #373, opened 2021-04-21
Fix broken PDF layout

Holzhaus:
These figure attributes completely broke the PDF layout (on *all* subsequent pages)

Left: Before this patch
Right: After this patch

Screenshot from 2021-04-21 15-04-33


Holzhaus:
@Swiftb0y @ronso0 merge? this is a prerequisite for 2.3 IMHO, because the PDFs are completely broken without it.

ronso0:
Right now I don't have to time to build and compare the PDFs, sorry.
This is a minor fix and if your 100% sure it's fixed you can merge it yourself IMO.

PR #377, opened 2021-04-30
Update source translations and Transifex configuration.

Holzhaus:
This is overdue. I didn't run `tx pull -s` or pulled in updated translations because I have no API access to Transifex.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -139,3 +139,849 @@ minimum_perc = 0
 source_file = source/locale/pot/chapters/vinyl_control.pot
 source_lang = en_US
 type = PO
+
+[mixxxdj-manual.404]
+file_filter = source/locale/<lang>/LC_MESSAGES/404.po
+source_file = source/locale/pot/404.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.glossary_]
+file_filter = source/locale/<lang>/LC_MESSAGES/glossary.po
+source_file = source/locale/pot/glossary.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.index]
+file_filter = source/locale/<lang>/LC_MESSAGES/index.po
+source_file = source/locale/pot/index.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.index_developer]
+file_filter = source/locale/<lang>/LC_MESSAGES/index_developer.po
+source_file = source/locale/pot/index_developer.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.manual_guidelines]
+file_filter = source/locale/<lang>/LC_MESSAGES/manual_guidelines.po
+source_file = source/locale/pot/manual_guidelines.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.sphinx]
+file_filter = source/locale/<lang>/LC_MESSAGES/sphinx.po
+source_file = source/locale/pot/sphinx.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.todolist]
+file_filter = source/locale/<lang>/LC_MESSAGES/todolist.po
+source_file = source/locale/pot/todolist.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.chapters--advanced_topics]
+file_filter = source/locale/<lang>/LC_MESSAGES/chapters/advanced_topics.po
+source_file = source/locale/pot/chapters/advanced_topics.pot
+source_lang = en
+type = PO
+
+[mixxxdj-manual.chapters--appendix]
+file_filter = source/locale/<lang>/LC_MESSAGES/chapters/appendix.po
+source_file = source/locale/pot/chapters/appendix.pot
+source_lang = en
+type = PO
+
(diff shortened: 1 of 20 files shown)


esbrandt:
How to cope with the sheer amount of new resources file (100+)?
The current page on Transifex already looks packed.

Screenshot_2021-05-02 The Mixxx DJ manual translation project on Transifex

At least we need some kind of organization by branch, similar to what we do with the app translation resources.

Holzhaus:
@esbrandt sure. I'm not familiar with transifex. Can you take care of that?

esbrandt:
One possible way of organizing is to generate just one single POT file, similar we do for the app with ``mixxx.ts``. 
https://github.com/mixxxdj/manual/blob/796174026e1f7af446e5b69a570c64a3af38f51f/source/conf.py#L116-L120

``gettext_compact = "Mixxx User Manual"``

It generates a single humongous ``*.pot`` file with a 100+k word count. In my brief test outside the Mixxx organization on Transifex, this file was successfully parsed. Translation memory filled up existing translations.

But there is additional work necessary on Transifex. Since we heavily rely on custom roles in the code (e.g. ``:guilabel:`` and ``:mixxx:coref:``), we need to add a custom translation and make sure they are left untouched in the translation.

If there is consensus how to proceed, I'll add the checks. As a safety measure, there should be a test run within the Mixxx organization on Transifex pre-merge.

Holzhaus:
> One possible way of organizing is to generate just one single POT file, similar we do for the app with `mixxx.ts`.

Is there no other way to work with "branches"? Using a single POT file will make it hard to focus on the more important parts, e.g. translating main manual content revelant to all users vs. translating controller mapping documentation.

> But there is additional work necessary on Transifex. Since we heavily rely on custom roles in the code (e.g. :guilabel: and :mixxx:coref: ), we need to add a custom translation check and make sure they are left untouched in the translation.

This is also true for built-in roles. I think I saw sphinx errors when building the french version of the manual because `:term:` was translated to `:terme:` and stuff like that.

Holzhaus:
@esbrandt I'll close this for now, can you take care of this? My experience with Transifex is virtually nonexistent.

esbrandt:
> > One possible way of organizing is to generate just one single POT file, similar we do for the app with `mixxx.ts`.
> 
> Is there no other way to work with "branches"? Using a single POT file will make it hard to focus on the more important parts, e.g. translating main manual content revelant to all users vs. translating controller mapping documentation.

To my knowledge, Transifex does not support versioning and branching. All I have found is [this](https://docs.transifex.com/localization-tips-workflows/how-to-work-with-multiple-branches#section-introduction). To my understanding, that does not exactly fit our workflow.

What we can do now:
   * Leave it as it is: Dump all new resources according to this PR.
      We can categorize (tag) resources, give them a descriptive name, add custom translation checks, and start translating.

   * Create a project on Transifex per git branch. And continue from there like above. I do not think this works well in the long term.

Holzhaus:
> * Leave it as it is: Dump all new resources according to this PR.
>   We can categorize (tag) resources, give them a descriptive name, add custom translation checks, and start translating.

Short term this probably makes sense.

Long-term, we can check if we can use an alternative to transifex that does support branching (migration should be pretty easy with po(t) files I guess). For example, [Crowdin seems to support branches](https://support.crowdin.com/versions-management/#branches) and has a [free open-source plan](https://crowdin.com/pricing#open-source-plans). Downside: We lose our existing userbase on transifex. We should evaluate if Crowdin or some other service fits our needs better than transifex. But we shouldn't rush this now.

Holzhaus:
We can discuss alternatives to Transifex on Zulip: https://mixxx.zulipchat.com/#narrow/stream/109176-documentation/topic/Migrating.20away.20from.20Transifex.3F

Holzhaus:
@esbrandt Apparently there *is* branch support on transifex:

```
$ tx push --help
usage: tx [-h] [-l LANGUAGES] [-r RESOURCES] [-f] [--skip] [-s] [-t] [--no-interactive] [-x] [--use-git-timestamps] [-b [BRANCH]]
          [--parallel]

This command pushes all local files that have been added to Transifex to the remote server. All new translations are merged with
existing ones and if a language doesn't exists then it gets created. If you want to push the source file as well (either because
this is your first time running the client or because you just have updated with new entries), use the -f|--force option. By
default, this command will push all files which are watched by Transifex but you can filter this per resource or/and language.

optional arguments:
  -h, --help            show this help message and exit
[...]
  --use-git-timestamps  Compare local files to their Transifex version by their latest commit timestamps. Use this option, for
                        example, when cloning a Git repository.
  -b [BRANCH], --branch [BRANCH]
                        Pull for a specific branch. Default is currentbranch if exists.
  [...]
```

On transifex, it seems like the manual translations always use the `release-1.12.0` category?! Maybe we were just using it incorrectly.

I'm really unfamiliar with transifex, it would be great if you could take care of syncing the manual translation for the 2.3.

esbrandt:
From reading the FAQ, I understand, that the ``--branch`` switch will just dump the same resources under another slug (name) on Transifex, thus doubling the total number of resources visible. We already use different slugs for the dev and stable releases. Nonetheless, I’ll have a look today.

PR #382, opened 2021-05-06
Move mixxx controls to Appendix

ronso0:


ronso0:
while I'm at it, any objections regarding

> Btw, in the Appendix, **16.1.** Additional Resources should IMO be moved to the end > so it's **16.5** and then there's only **16.6** Version History after that.

ronso0:
can't find the reason that makes Sphinx / Build check fail https://github.com/mixxxdj/manual/pull/382/checks?check_run_id=2518202803
Only see some Title warnings..

Holzhaus:
> can't find the reason that makes Sphinx / Build check fail https://github.com/mixxxdj/manual/pull/382/checks?check_run_id=2518202803
> Only see some Title warnings..

    Title level inconsistent

The title level is defined by the characters.


Here's an example:

    Level 1
    =======

    Level 2
    -------

    Level 3
    ~~~~~~~

    Level 2 again
    -------------

    What level is this?
    ^^^^^^^^^^^^^^^^^^^

The last headline will cause such a warning. It should use `~~~` instead if it's supposed to be on level 3, or `---` or for level 2, or `===` for level 1. Level 2 followed by level 4 without level 3 in between is not possible.

Holzhaus:
Btw, I think we should split the appendix into different files and use include directives. Otherwise we still have the same problem of endless scrolling when editing the RST file.

Holzhaus (approved):
Thank you!

PR #385, opened 2021-05-09
pre-commit: Optimize SVGs automatically

Holzhaus:
It get annoying to remind people to optimize SVGs using Inkscape... and apparently that isn't even an option on WIndows.


tools/optimize_svg.py
```diff
@@ -0,0 +1,35 @@
...
+
+ARGS = (
+    "--enable-comment-stripping",
+    "--enable-id-stripping",
+    "--enable-viewboxing",
+    "--indent=none",
+    "--no-line-breaks",
+    "--remove-descriptions",
+    "--remove-metadata",
+    "--remove-titles",
```
ronso0, line 15: all done by --remove-descriptive-elements:
```
    --remove-descriptive-elements
                                 remove <title>, <desc> and <metadata>
                                 elements

```

Swiftb0y (approved):

Swiftb0y:
@Holzhaus @ronso0 do you think it would make sense to add the same pre-commit hook to the main repo for minimizing the graphics used in skins?

Holzhaus:
We can. Maybe I'll make a separate repo so that we can avoid duplicating the code in each repo.

PS: I wish scour supported conversion into SVG Tiny. That would solve a lot of headaches.

ronso0:
I agree with @esbrandt 
In existing skins most graphics are optimized already, and the maintainer takes care of new icons.
For new skins we need to request it only once (not repeatedly like for mapping documentations).

Also, with the Qt skins the view in the vector program is not necessarily WYSIWYG as soon as gradients and transparencies are used. then we cannot rely on the script unfortunately, icons needs to be checked manually.

PR #393, opened 2021-05-15
Inclusive language

Holzhaus:
Following up on our commitment we made here: https://mixxx.org/news/2020-06-29-black-lives-matter/

This replaces the term "master" with "main", "master sync" with "sync lock", "sync master" with "sync leader". I did not replace occurences where I had the impression that it referred to a hardware label, as this would make it harder to find the corresponding button/fader.

This also fixes an actual issue because we already replaced the term in the GUI, but the old term was still used in the docs.


Holzhaus:
Oops, this should go into main, we didn't rename the outputs in 2.3 yet.

ronso0:
> Oops, this should go into main, we didn't rename the outputs in 2.3 yet.

hmm.. It's also rather inconsistent for the GUI documentation and in skins:
we use MAIN in LateNight (only) now, also in the [mixer chapter](https://manual.mixxx.org/2.3/en/chapters/user_interface.html#the-mixer-section) but the [Master],gain tooltip still says Master.

I think we should add this to 2.3 --except maybe the sound hardware preferences documentation-- but the remaining skin labels can easily be adjusted for 2.3 (even with the label/tooltip inconsistency).

Holzhaus:
Hmm, I think it would be confusing to talk about "Main output" if it's named "master output" in the preferences. I'm not particularily keen on splitting up this issue. If you think it's needed, please go ahead, otherwise let's just postpone it until 2.4.

ronso0:
ah okay, I see https://github.com/mixxxdj/mixxx/pull/3868 was merged to main because of String freeze.

Side note: why should Master/Main be translated at all? I vaguely remember we discussed not to translate terms that are virtually omnipresent hardware labels on most DJ/audio gear.

Holzhaus:
So, should we merge this into main?

source/glossary.rst
```diff
@@ -20,10 +20,11 @@ Glossary of Terms
 
    crossfader
      The crossfader is a slider that determines how much each deck of audio
-     contributes to the :term:`master output`.
+     contributes to the :term:`main output`.
 
+   main output
    master output
```
Swiftb0y, line 26: maybe emphasize that the term "master output" has been used historically, but is being phased out so people will know which one to use?

source/hardware/controllers/denon_mc7000.rst
```diff
@@ -308,7 +308,7 @@ LEDs
 
 The Channel Volume Meters matches to the ones shown in Mixxx GUI. Only when clipping the red LED illuminates.
 
-The Master Volume Meter is not correlated to Mixxx GUI as the controller handles that in Hardware.
+The Main Volume Meter is not correlated to Mixxx GUI as the controller handles that in Hardware.
```
Swiftb0y, line 311: I'm unsure whether we should change that since "Main Volume Meter" seems to be referring to the meter which is in the region labelled "master" on the controller. 
image

Swiftb0y (commented):
Thanks, two small nitpicks, LGTM otherwise.

source/hardware/controllers/denon_mc7000.rst
Holzhaus: That depends on the device.

Swiftb0y: I fully agree that its device dependent (I was actually positively surprised seeing how much hardware was already using main instead of master). However, the screenshot I posted was from the controller the manual page is on. So I was proposing using Master, since thats whats labelled on the controller.

Holzhaus: Ah I overlooked that. Okay, we can put it in a hwlabel to emphasize that it's the label on the device.

source/glossary.rst
Holzhaus: Can you suggest a wording? I'm a bit uncreative today.

Swiftb0y: Maybe add an extra sentence at the end of the description like this:
> Please note: the term "master output" has been used historically but is in the [process of being phased out](https://mixxx.org/news/2020-06-29-black-lives-matter/). Prefer using "main output".

I'm not too good at writing either so feel free to make changes as you see fit

Swiftb0y (approved):
Thank you very much LGTM

Swiftb0y:
I guess owen doesn't have any interest / time to review this so I'll go ahead and merge now.

ywwg:
sorry I didn't have time to review.  thanks for doing the work

PR #398, opened 2021-06-23
Remove obsolete _redirects file

Holzhaus:
This file is written by the build_html.sh script.


daschuer:
Another broken link :-/
Do we need to add this file to .gitignore that it does not re-appear from a users worktree?

Holzhaus:
It won't, it's only written by the build script into the output directory which is already ignored.

Holzhaus:
Merge?

daschuer (approved):
Yes, thank you.

PR #407, opened 2021-07-05
Use sphinx-notfound-page to somewhat fix #395

### LINKED ISSUE #395: page-not-found not styled
see:
https://manual.mixxx.org/2.3/en/chapters/hardware/controllers/#page-not-found

ICTman1076:
This PR will not fix the issue entirely (see problems below), but is intended as a starting point for further fixes. I am not familiar with sphinx so there may be something I've missed.

In this current state, the language links on the 404 page do not work, and it will always use the English 404 page, but styling and other links should be fixed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,5 @@ transifex-client
 sphinxcontrib-svg2pdfconverter
 sphinx-rtd-theme==0.5.2
 sphinx-multiversion==0.2.1
+sphinx-notfound-page
 ./sphinx-mixxx
--- a/source/conf.py
+++ b/source/conf.py
@@ -36,6 +36,7 @@ extensions = [
     "sphinx_rtd_theme",
     "sphinx_multiversion",
     "sphinx_mixxx",
+    "notfound.extension",
 ]
 
 todo_include_todos = True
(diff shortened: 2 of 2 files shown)


Holzhaus:
Thanks for contributing. However, we already have a 404 page. The issue is that the urls to images/styles/etc are relative. So it works for https://manual.mixxx.org/2.3./en/foo but not for https://manual.mixxx.org/2.3/en/foo/bar. We probably just need to override [`html_baseurl`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl).

ICTman1076:
> Thanks for contributing. However, we already have a 404 page. The issue is that the urls to images/styles/etc are relative. So it works for https://manual.mixxx.org/2.3./en/foo but not for https://manual.mixxx.org/2.3/en/foo/bar. We probably just need to override [`html_baseurl`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl).

This is what this PR resolves. The `sphinx-notfound-page` plugin automatically takes the existing `404.rst` and makes all its links absolute. The only problem is the language links, which it does not correctly fix.

However, if that solution works then it would avoid having to install another pip module, which would be useful. But I did a little research and this appeared to be one of the best solutions for a 404 page, [used by Read the Docs and others](https://sphinx-notfound-page.readthedocs.io/en/latest/who-is-using-it.html).

Holzhaus:
Hmm, it doesn't work according to the Deploy preview:

Holzhaus:
Because of our versioning system, we always build from the main branch. I'll retarget on main, then it might work.

ronso0:
@ICTman1076 
Could you please rebase onto `main` branch and force-push?

PR #417, opened 2021-07-27
 Add entries for replaygain update functionality.

ywwg:
Also correct two headings that don't match the UI wording.

https://github.com/mixxxdj/mixxx/pull/4031

### DIFF
--- a/source/chapters/appendix/mixxx_controls.rst
+++ b/source/chapters/appendix/mixxx_controls.rst
@@ -2516,8 +2516,8 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
   value for the deck will be centered so that there is no audible difference in
   track volume, so this operation is safe to use during performance, if the controller mapping uses soft-takeover for the pregain knob.
 
-   :range: binary
-   :feedback: ReplayGain value is updated in library, deck pregain is reset to 1.0.
+  :range: binary
+  :feedback: ReplayGain value is updated in library, deck pregain is reset to 1.0.
 
 
 .. mixxx:control:: [ChannelN],vinylcontrol_enabled
--- a/source/chapters/appendix/mixxx_controls.rst
+++ b/source/chapters/appendix/mixxx_controls.rst
@@ -2506,6 +2506,19 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
 
    .. versionadded:: 2.0.0
 
+.. mixxx:control:: [ChannelN],update_replaygain_from_pregain
+                   [PreviewDeckN],update_replaygain_from_pregain
+                   [SamplerN],update_replaygain_from_pregain
+
+  Applies the deck pregain knob value to the detected ReplayGain value for the
+  current track. This is a way to update the ReplayGain value of a track if it
+  has been detected incorrectly. When this option is selected, the pregain value
+  for the deck will be adjusted so that there is no audible difference in track
+  volume, so this operation is safe to use during performance.
+
+   :range: binary
+   :feedback: ReplayGain value is updated in library, deck pregain is adjusted.
+
 
 .. mixxx:control:: [ChannelN],vinylcontrol_enabled
                    [PreviewDeckN],vinylcontrol_enabled
--- a/source/chapters/appendix/mixxx_controls.rst
+++ b/source/chapters/appendix/mixxx_controls.rst
@@ -2512,12 +2512,12 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
 
   Applies the deck pregain knob value to the detected ReplayGain value for the
   current track. This is a way to update the ReplayGain value of a track if it
-  has been detected incorrectly. When this option is selected, the pregain value
-  for the deck will be adjusted so that there is no audible difference in track
-  volume, so this operation is safe to use during performance.
+  has been detected incorrectly. When this control is triggered, the pregain
+  value for the deck will be centered so that there is no audible difference in
+  track volume, so this operation is safe to use during performance.
 
    :range: binary
-   :feedback: ReplayGain value is updated in library, deck pregain is adjusted.
+   :feedback: ReplayGain value is updated in library, deck pregain is reset to 1.0.
 
 
 .. mixxx:control:: [ChannelN],vinylcontrol_enabled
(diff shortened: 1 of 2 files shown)


source/chapters/appendix/mixxx_controls.rst
```diff
@@ -2506,6 +2506,19 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
...
 
+.. mixxx:control:: [ChannelN],update_replaygain_from_pregain
+                   [PreviewDeckN],update_replaygain_from_pregain
+                   [SamplerN],update_replaygain_from_pregain
+
+  Applies the deck pregain knob value to the detected ReplayGain value for the
+  current track. This is a way to update the ReplayGain value of a track if it
+  has been detected incorrectly. When this option is selected, the pregain value
+  for the deck will be adjusted so that there is no audible difference in track
+  volume, so this operation is safe to use during performance.
```
ronso0, line 2517:
```suggestion
  has been detected incorrectly. When this control is triggered, the pregain value
  for the deck will be centered so that there is no audible difference in track
  volume, so this operation is safe to use during performance.
```

source/chapters/appendix/mixxx_controls.rst
```diff
@@ -2506,6 +2506,19 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
...
+
+   :range: binary
+   :feedback: ReplayGain value is updated in library, deck pregain is adjusted.
```
ronso0, line 2520:
```suggestion
   :feedback: ReplayGain value is updated in library, deck pregain is centered.
```

ronso0 (changes requested):

source/chapters/appendix/mixxx_controls.rst
ywwg: this is not quite true -- the deck that was the source of the adjustment will end up at the center position, but if there are any other decks with the track loaded they may not be adjusted to the center.

ywwg: But for the sake of simplicity, maybe it's fine to elide that detail :)

source/chapters/appendix/mixxx_controls.rst
ywwg: applied manually

source/chapters/appendix/mixxx_controls.rst
ywwg: applied manually

source/chapters/appendix/mixxx_controls.rst
ronso0: ah okay, I just edited to "...deck pregain is reset to 1.0"

source/chapters/appendix/mixxx_controls.rst
ronso0: ah, good point.

source/chapters/appendix/mixxx_controls.rst
ywwg: done

ronso0: thx!

source/chapters/appendix/mixxx_controls.rst
```diff
@@ -2506,6 +2506,19 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
...
+                   [SamplerN],update_replaygain_from_pregain
+
+  Applies the deck pregain knob value to the detected ReplayGain value for the
+  current track. This is a way to update the ReplayGain value of a track if it
+  has been detected incorrectly. When this control is triggered, the pregain
+  value for the deck will be centered so that there is no audible difference in
+  track volume, so this operation is safe to use during performance.
+
+   :range: binary
+   :feedback: ReplayGain value is updated in library, deck pregain is reset to 1.0.
```
ronso0, line 2520: "One more thing..." ; )
```suggestion
  :range: binary
  :feedback: ReplayGain value is updated in library, deck pregain is reset to 1.0.
```
to align it with the above paragraph, like other controls

ronso0 (changes requested):

source/chapters/appendix/mixxx_controls.rst
```diff
@@ -2506,6 +2506,19 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
...
 
+.. mixxx:control:: [ChannelN],update_replaygain_from_pregain
+                   [PreviewDeckN],update_replaygain_from_pregain
+                   [SamplerN],update_replaygain_from_pregain
+
+  Applies the deck pregain knob value to the detected ReplayGain value for the
+  current track. This is a way to update the ReplayGain value of a track if it
+  has been detected incorrectly. When this control is triggered, the pregain
+  value for the deck will be centered so that there is no audible difference in
+  track volume, so this operation is safe to use during performance.
```
JoergAtGithub, line 2517:
```suggestion
  track volume, so this operation is safe to use during performance, if the controller mapping uses soft-takeover for the pregain knob.
```

Holzhaus: Hmm, this always applies. Otherwise even adjusting the gain knob via mouse will lead to issues, regardless of replay gain. So I IMHO this addition is not really needed.

ronso0: This is a long sentence already. Also, a controller is not necessarily the default.
What about
"If you use a controller make sure to enable SoftTakeover for the pregain knob for full comfort"

ronso0: > Hmm, this always applies. Otherwise even adjusting the gain knob via mouse will lead to issues, regardless of replay gain. So I IMHO this addition is not really needed.

true. let's leave it as is.

JoergAtGithub: This is the documentation page for developers of controller mappings. And this is the point, which needs to be considered in a mapping, because all other COs can be used safely without implemnting soft-takeover, but this one not.

ywwg: I think it's fine to add this clarification

Holzhaus (approved):

ronso0:
LGTM!

PR #423, opened 2021-08-20
Update links to macOS software

ferdnyc:
**Disclaimer:** I'm not a Mac user, and don't have access to one. So, not only can I not guarantee that the software linked here is _working correctly_, but the links haven't been tested **at all** beyond verifying that they're not 404s and appear to reach the expected tools.

I noticed that some of the links in the "Effects via AU Lab on macOS" section were quite moldy.
1. The SoundFlower link pointed to a repo that hasn't been updated since 2013. The replacement is a newer fork that was at least updated in 2020. It mentions still-supported tool versions when discussing compatibility (as opposed to "XCode 4" or "XCode 5", in the previous repo), so that seems like a win.
2. The AU Lab link pointed to a 404 page, but Apple still has a page on their site (https://www.apple.com/apple-music/apple-digital-masters/) which links to a working download.

### DIFF
--- a/source/chapters/advanced_topics.rst
+++ b/source/chapters/advanced_topics.rst
@@ -171,10 +171,10 @@ Effects via AU Lab on macOS
 On macOS there is a simple and free way to give Mixxx access to the
 collection of AU/VST/MAS plugins that are installed on your system.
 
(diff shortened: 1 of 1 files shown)


Be-ing:
thanks

Be-ing:
FWIW we will be building PortAudio with the JACK backend enabled on macOS for Mixxx 2.4.

PR #430, opened 2021-09-18
Update korg_kaoss_dj.rst

raphaelquast:
add new assignments for
https://github.com/mixxxdj/mixxx/pull/4301

### DIFF
--- a/source/hardware/controllers/korg_kaoss_dj.rst
+++ b/source/hardware/controllers/korg_kaoss_dj.rst
@@ -62,21 +62,21 @@ Mapping
 +------------------------------------------------+------------------------------------------------+------------------------------------------------+------------------------------------------------+
 | 12                                             | Touch Slider mode button                       | Switches between the three touch slider mode   |                                                |
 +------------------------------------------------+------------------------------------------------+------------------------------------------------+------------------------------------------------+
-| 13                                             | Touch Slider (Normal mode)                     | Left Side - Nudges (pitch -)                   | Moves to the specified position in the song    |
+| 13                                             | Touch Slider (Normal mode)                     | Left - Nudges (pitch -)                        | Moves to the specified position in the song    |
 |                                                |                                                | Center - Enables the touch wheel's Scratch     | (slider)                                       |
 |                                                |                                                | mode                                           |                                                |
-|                                                |                                                | Right Side - Nudges (pitch +)                  |                                                |
+|                                                |                                                | Right - Nudges (pitch +)                       |                                                |
 +------------------------------------------------+------------------------------------------------+------------------------------------------------+------------------------------------------------+
-|                                                | Touch Slider (Hot Cue mode)                    | Left Side - Sets Hot Cue 1, and moves to Hot   | Deletes Hot Cue 1                              |
-|                                                |                                                | Cue 1                                          | Deletes Hot Cue 2                              |
-|                                                |                                                | Center - Sets Hot Cue 2, and moves to Hot Cue  | Deletes Hot Cue 3                              |
+|                                                | Touch Slider (Hot Cue mode)                    | Left - Sets Hot Cue 1, and moves to Hot        | Left - Deletes Hot Cue 1                       |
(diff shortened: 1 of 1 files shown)


source/hardware/controllers/korg_kaoss_dj.rst
```diff
@@ -42,15 +42,15 @@ Mapping
...
-| 10                                             | Touchpad (controller mode)                     | Controls the effects of the DJ software. The   | Adjusts the Beats Multiplier (TODO)            |
-|                                                |                                                | vertical axis controls the mix (dry/wet) knob  |                                                |
-|                                                |                                                | and the horizontal axis controls the super     |                                                |
+| 10                                             | Touchpad (controller mode)                     | Controls the effects of the DJ software. The   | Controls Quick-Effect super knobs              |
+|                                                |                                                | vertical axis controls the mix (dry/wet) knob  | the horizontal-axis controls deck 1            |
+|                                                |                                                | and the horizontal axis controls the super     | the vertical-axis controls deck 2              |
```
ronso0, line 53: If you're touching this unrelated lines, please try to boil down the text
`Controls the effects of the DJ software. The vertical axis controls the mix (dry/wet) knob and the horizontal axis controls the super knob.`
to something like
```
Controls effect unit 1.
Y axis: Mix knob (dry/wet)
X axis: Super knob.
```

source/hardware/controllers/korg_kaoss_dj.rst
```diff
@@ -62,29 +62,30 @@ Mapping
 +------------------------------------------------+------------------------------------------------+------------------------------------------------+------------------------------------------------+
 | 12                                             | Touch Slider mode button                       | Switches between the three touch slider mode   |                                                |
 +------------------------------------------------+------------------------------------------------+------------------------------------------------+------------------------------------------------+
-| 13                                             | Touch Slider (Normal mode)                     | Left Side - Nudges (pitch -)                   | Moves to the specified position in the song    |
+| 13                                             | Touch Slider (Normal mode)                     | Left - Nudges (pitch -)                        | Moves to the specified position in the song    |
```
ronso0, line 65: this and the rows below:
I think some formatting or line breaks in  columns 3+4 would improve readability a lot.
Current state looks like this:
image

The text can also be improved, for example
```
Set hotcue / go to set hotcue
Left: 1 / Center: 2 / Right: 3
```
and in a similiar way for the shifted mode description.

raphaelquast: the text is in fact formatted with line-breaks like this:
```
| ... | ... | Left - Deletes Hot Cue 1  |
| ... | ... | Center - Deletes Hot Cue 2|
| ... | ... | Right- Deletes Hot Cue 3  |
```

It should be formatted in the doc (parsed via Sphinx i guess) as 3 individual lines.
However Github's preview interprets this as a Markdown table that does not support newlines which is why it looks like a mess in your screenshot 😄

ronso0: this is not just a preview, it's the actual online manual.
for markdown we need better separators then to increase readability.

raphaelquast: hmm, well as far as I know this is supposed to be the correct way for specifying newlines in re-structured text... 
I think what's happening here is that the lines exceed the max. number of characters and then auto-wrapping takes over and the newlines are no longer correctly parsed...

this would explain this:  (from the online-doc)  
(all entries are written with newlines but it only works as intended for "Deletes Hot Cue ."
image

ronso0: yes, I understand that.
what I was proposing as "better separators" is simply something like `Left: set hotcue... ` instead of `Left - set hotcue... `

raphaelquast: ok, yep that seems reasonable! 
I'll  try to come up with a better readable version as soon as i find the time!

PR #435, opened 2021-09-29
Behringer BCR2000: Fix download links

git-developer:
The links to static files work in the netlify preview, but are broken on the official manual because _static is missing.


Swiftb0y (approved):
Thanks

PR #443, opened 2021-10-07
Keyboard table: reformat to 3 colums, split into categories

ronso0:
* make functions are easier to spot with tables per category
* separate tables only need 3 columns which avoids quite a few lines breaks

[now](https://manual.mixxx.org/2.3/en/chapters/appendix/keyboard_mapping_table.html#appendix-keyboard) vs. [new](https://deploy-preview-443--mixxx-manual.netlify.app/chapters/appendix/keyboard_mapping_table.html)


Swiftb0y (approved):
Thanks, those are valid arguments. I don't see a reason to object to merging this (whether its duplicated/redundant or not).

PR #446, opened 2021-10-14
controls: complete '[Library],GoToItem' description

ronso0:
list all context-specific actions triggered by `[Library],GoToItem`


source/chapters/appendix/mixxx_controls.rst
```diff
@@ -3183,7 +3183,14 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
 
 .. mixxx:control:: [Library],GoToItem
 
-   Equivalent to double clicking the currently selected item
+   Triggers different actions, deepending on which interface element currently has keyboard focus:
```
Holzhaus, line 3186:
```suggestion
   Triggers different actions, depending on which interface element currently has keyboard focus:
```

source/chapters/appendix/mixxx_controls.rst
```diff
@@ -3183,7 +3183,14 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
...
+
+   | **Search bar** text box: moves focus to tracks table
```
Holzhaus, line 3188: The line break preservation is hacky. This should either be a definition list or an enumeration (unordered list) IMHO.

Another alternative would be a two-column table.

ronso0: hmm neither really fits IMO.
I'll make it a table
image

Holzhaus (approved):
LGTM, thank you.

PR #452, opened 2021-11-09
document [Library],focused_widget controls

ronso0:


source/chapters/appendix/mixxx_controls.rst
```diff
@@ -3425,6 +3425,33 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
    .. versionadded:: 2.1.0
 
 
+.. mixxx:control:: [Library],focused_widget
+
+   Read this control to know which library widget is currently focused, or write in order to focus a specific library widget.
+
+   This control can be used in controller scripts to trigger context-specific actions. For example, if the
+   tracks table has focus, pressing a button loads the selected track to a specific deck, while the same
+   button would clear the search if the search bar is focused.
```
Swiftb0y, line 3434: not sure if it makes sense to mention this in the CO description. I don't think we have usage examples in other CO docs.

ronso0: hmm, true. Though the only intention (and use case IMO) for this CO is scripting, that's why I think it deserves an example.

Swiftb0y: valid point

Swiftb0y (approved):

PR #453, opened 2021-11-09
Double click knobs/faders to reset

ronso0:


source/chapters/controlling_mixxx.rst
```diff
@@ -22,6 +22,8 @@ can perform actions by pointing and clicking with your mouse.
   default action for that control. Moreover if you click on a value in the track
   list of the Mixxx :ref:`library <library-interface>`, you can change the
   value. For example the year of a track.
+* **Double Left Click**: Double-click on any knob or fader to reset it to its
+  default value.
```
Swiftb0y, line 26: maybe mention that this is useful when interacting with mixxx on a touchscreen?

ronso0: hmm,.. why do you think this is worth mentioning?

Swiftb0y: I don't know if people understand that they can reset a control by double tapping on their screen when reading it. Maybe I'm underestimating our users. 
It also helps ot explain why we have two different ways to reset the control (right click being more intuitive for mouse users and double click being more intuitive for touchpad/-screen users).

ronso0: okay, I'll add the the touch screen hint.

ronso0: **Double Left Click**: Double-click on any knob or fader to reset it to its
  default value. Especially with touch screens this method is easier and faster  
  than right-clicking which usually requires long-pressing the screen.

Alright?

Swiftb0y: well long-pressing the screen is not even an option for faders and knobs since that is just moving the control. What about 
> Double Left Click: Double-click on any knob or fader to reset it to its default value. This is especially useful on touchscreens as executing a right-click on them is usually tedious.

ronso0: :+1:

Swiftb0y (approved):
Thanks

PR #454, opened 2021-11-12
fix macOS settings paths

ronso0:


ronso0:
The paths are used in a few places. Is there a way to have a global var to ease maintenance?

Be-ing:
thanks

PR #457, opened 2021-11-28
Generate patch file artifact

JoergAtGithub:
Generate patch file artifact on GitHub, with the pre-commit changes, as on mixxxdj/mixxx

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,3 +13,18 @@ jobs:
     - uses: pre-commit/action@v2.0.0
       env:
         SKIP: rstcheck
+
+    - name: "Generate patch file"
+      if: failure()
+      run: |
+        git diff-index -p HEAD > "${PATCH_FILE}"
+        [ -s "${PATCH_FILE}" ] && echo "UPLOAD_PATCH_FILE=${PATCH_FILE}" >> "${GITHUB_ENV}"
+      env:
+        PATCH_FILE: pre-commit.patch
+
+    - name: "Upload patch artifact"
+      if: failure() && env.UPLOAD_PATCH_FILE != null
+      uses: actions/upload-artifact@v2
+      with:
+        name: ${{ env.UPLOAD_PATCH_FILE }}
+        path: ${{ env.UPLOAD_PATCH_FILE }}


.github/workflows/pre-commit.yml
```diff
@@ -13,3 +13,18 @@ jobs:
     - uses: pre-commit/action@v2.0.0
       env:
         SKIP: rstcheck
+
+    - name: "Generate patch file"
```
daschuer, line 17:
```suggestion
    - name: "Generate pre-commit.patch"
```

.github/workflows/pre-commit.yml
```diff
@@ -13,3 +13,18 @@ jobs:
...
+      if: failure()
+      run: |
+        git diff-index -p HEAD > "${PATCH_FILE}"
+        [ -s "${PATCH_FILE}" ] && echo "UPLOAD_PATCH_FILE=${PATCH_FILE}" >> "${GITHUB_ENV}"
```
daschuer, line 21: What does [ -s ... ] do?
I think it is anyways much more straight forward to define common environment variables in the header section if the whole file.

daschuer (changes requested):
Thank you for this little helper.
I have left some comments for improvement.

.github/workflows/pre-commit.yml
Holzhaus: `-s` is true if the file exists and is not empty. This ensures that the patch file is only uploaded when necessary.

.github/workflows/pre-commit.yml
Holzhaus: This is part of the pre commit workflow, so the original title is fine IMHO. I'd also prefer to keep the workflow consistent across all repos.

Holzhaus (approved):

.github/workflows/pre-commit.yml
daschuer: Ok, It was just an idea for improvement.

.github/workflows/pre-commit.yml
Holzhaus: > Oh I have finally found it.
> It was as hard to find because google is not good in finding a single letter expression:
> https://tldp.org/LDP/abs/html/fto.html

Next time, you can type `man test`, then `/-s` 🙂

> I think we can remove this hard to read code anyway by a global environment variables for the patch name in the header section. Even without the patch the upload is skipped if the patch file dies not exist alreadautomatically shipped if the patch file dies not exist. 

This is not equivalent. This check ensures that the file is also not empty (because it's possible that pre-commit fails but is unable to make automated changes).

daschuer (approved):

daschuer:
Thank you for your explanations. LGTM

PR #458, opened 2021-11-29
 Run make gettext and add new/modified files

ywwg:


ywwg:
There are a whole bunch of broken links in the manual... should I fix / remove those here or can we address it later?  (I'd prefer to do this incrementally)

Swiftb0y:
Most of those seem to be false positives. Either because some conncections actually seem to block us (403) or because there is some subtle TLS error. All of the links do actually work in the browser.

ywwg:
(alternatively, we can ignore it for this PR and file a bug)

Swiftb0y:
More importantly though, who can review this PR? I've never ran this command so I assume it just updates the translation files? So we can basically merge without looking at them, right?

Holzhaus:
IIRC @esbrandt is the translations expert.

daschuer:
Here is the guide for our mixxx repro. 
https://github.com/mixxxdj/mixxx/wiki/internationalization

I think the same workflow applies here as well. 
We normally push the changes directly to transiflex and into our repro. 
No need for a PR. 

I can do this here as well. And we can close this PR. 
Shall I?

ywwg:
I had to run this command to get the manual to pick up my new controller file, and it made all these other updates too

ywwg:
Is there something else I should have done in my other PR to get the build system to see my new file?

quote:

#### Update source translations

For every change to the manual source files (.rst) the source translation files
(.pot) must be re-generated. These are stored in `source/locale/pot` and contain
the text of every English phrase in the manual in a common format used for
translation and can be regenerated with:

    make gettext

daschuer:
Ah, yes I see the steps are described here: 
https://github.com/mixxxdj/manual#readme

I think you should do everything step by step.
This PR is only the first step. 

I think you can also push directly, too keep transiflex and the manual repository in sync. 

Do you have sufficient rights at transiflex? 

> Is there something else I should have done in my other PR to get the build system to see my new file?

You can already see your new file in the deploy preview:
https://deploy-preview-456--mixxx-manual.netlify.app/hardware/controllers/yaeltex_minimixxx.html 
I think this PR must be merged first, before you can add it here: 
https://github.com/mixxxdj/manual/blob/2.3/.tx/config

Holzhaus:
> I had to run this command to get the manual to pick up my new controller file, and it made all these other updates too

This is not necessary. With a clean build dir, the new controller documentation is included.

It probably makes sense to update the translations because they are old, but it's not necessary for building the manual.

ywwg:
ok, closing this

PR #461, opened 2021-12-05
Add entries to linkcheck ignore list for problematic URLs

ywwg:


source/conf.py
```diff
@@ -447,7 +447,19 @@
     # Ignore Twitter links because for some reason valid links sometimes fail
     # with "400 Bad Request"
     r"^https://twitter.com/\w+$",
+    # Ignore support links because we get 403 Client Errors (bad User Agent?)
+    r"^https://support.native-instruments.com/",
+    r"^https://support.serato.com/",
```
Swiftb0y, line 452:
```suggestion
    r"^https://support.native-instruments.com/.*",
    r"^https://support.serato.com/.*",
```
otherwise it won't match for the subdirectories I think. We have to test this.

source/conf.py
```diff
@@ -447,7 +447,19 @@
...
 ]
 
+linkcheck_ignore = [
+    # Unknown why this anchor doesn't appear to work -- could be js-generated.
+    r"^https://github.com/mixxxdj/mixxx/wiki/Midi-Scripting#",
+]
```
Swiftb0y, line 458: why are you adding the same option yet again (overwriting the old `linkcheck_ignore` I guess?)

ywwg: copy-paste error.

source/conf.py
ywwg: it's a partial regex match (no $ at the end of line), I did test and it works.

Swiftb0y: Great, I figured it doesn't work because pre-commit failed, but that was because the `linkcheck_ignore` list was overwritten.

Swiftb0y (approved):
LGTM, thanks

Swiftb0y (changes requested):
My bad, one thing. Would you mind squashing the second commit into the first since its just a fixup?

ywwg:
no problem, done

Swiftb0y (approved):
Thanks

Swiftb0y:
How do I apply this to 2.3 best? cherry-pick 7a52595d4ff306e9617bc10c68a75aaebc112092 and push directly?

PR #462, opened 2021-12-21
Update keyboard mapping table

ninomp:
* Add shortcut for opening track properties dialog
* Add shortcut for removing/hiding tracks
* Refer to Escape key as `Escape` instead of `ESC`

This is a follow-up to https://github.com/mixxxdj/mixxx/pull/4330 and https://github.com/mixxxdj/mixxx/pull/4347.


source/chapters/appendix/keyboard_mapping_table.rst
```diff
@@ -132,7 +132,11 @@ language settings the defaults may be different.
 +----------------------------------------+-----------------------------------------+------------------------------------------+
 | Clear search input                     | :kbd:`Ctrl` + :kbd:`Back`                                                          |
 +----------------------------------------+-----------------------------------------+------------------------------------------+
-| Exit search and leave focus            | :kbd:`ESC`                                                                         |
+| Exit search and leave focus            | :kbd:`Escape`                                                                      |
```
ronso0, line 135: while you're at it:
Ctrl + F is a global shortcut set in wtracktableview. cpp, so this should actually be
"Leave current widget and focus tracks table"

ronso0: Apart from that thos LGTM

ninomp: > "Leave current widget and focus tracks table"

This text does not fit in this table. Should I (just) enlarge the entire table?

ronso0: Well, just "Focus tracks table" is sufficient.

We may improve the documentaion later on, for example add `Tab`

ronso0:
LGTM, thank you!

PR #122, opened 2020-08-24
Controller Documentation

Holzhaus:

PR #464, opened 2022-01-01
Reordered structure of mapping chapters

JoergAtGithub:
This is a proposal for a clearer structure of the controller mapping related chapters:

- It contains two dummy chapters for the Wiki pages, which describe MIDI and HID mappings
- It moves unifys terms to supported and unsupported controllers
- It moves the topics into chapters with more meaningful names than Appendix and Advanced Topics

I think the new terms are much more user centric and less, what do you think?


source/chapters/how_to_map_controllers.rst
```diff
@@ -0,0 +1,23 @@
...
+
+With several dozens of DJ controllers supported out-of-the-box, Mixxx gives you
+comprehensive hardware control for your DJ mixes, see :ref:`control-midi`.
+
+Support for additional devices can be added to Mixxx by creating a new mapping
+file. This file tells Mixxx how to translate, or map, :term:`MIDI`/:term:`HID`
+messages from a controller into commands that Mixxx understands.
+
+You can download and share custom controller mappings in the
+`Mixxx User customizations forums <https://mixxx.org/forums/viewforum.php?f=6>`_.
```
ronso0, line 16: Instead of redirecting, let's use the new link right away
https://mixxx.discourse.group/c/controller-mappings/10

ronso0 (changes requested):

ronso0:
Thank you for starting to rework those chapters!

It makes sense to reorder and combine chapters for better discoverability.
It's a pity you already have invested quite some work -- I think we should discuss how the grouping and titles can be improved further:

Do we realy want to rename **_Hardware Manuals_** to _**Supported Controllers**_? That would group controllers and mixers? Do we need a distinction between the two? Atm we only support controllers but [@Holzhaus chose](https://github.com/mixxxdj/manual/pull/122#issuecomment-692786649) "Hardware" to "leave the door open for non-controller hardware, e.g. Soundcards, DVS hardware, etc."
For now I think the renaming is okay but we should consider other hardware when working on those chapters.

Regarding **_Supported Controllers_** and **_How to add support for new Controllers_**: I find the latter a rather clunky title (consider french and spanish maybe which usually require more and longer words, compared to english titles ;)
Also, I think we should work towards reducing the number of top-level chapters and improve the semantic order for better discoverability.
That said I recommend to combine those into _MIDI/HID Controllers_:
MIDI/HID Controllers
-Supported Controllers
-Add support for new Controllers

Then we can also add a third part (later on) _Change an existing mapping_. This a documentation topic I'm missing frequently when users ask for slight modifications of existing mappings and a manual chapter would be _very_ helpful

**_Advanced Topics_**
image
I think this section is now obsolete:
* _Additional Effects_ can be moved to _Effects_
* _Making a custom keyboard mapping_ can be moved to _Controlling Mixxx > Keyboard_
* _Deleting your library_ can either be removed or added to _Appedix > Mixxx settings directory_

Btw _Controlling Mixxx_ chapters should IMO be split into separate files so they don't appear on one page (which i consider confusing tbh)

What do you think?

I like the changes in general. The only thing I'd like to be implemented now is 
> I'd rather combine those into _MIDI/HID Controllers_.
MIDI/HID Controllers
-Supported Controllers
-Add support for new Controllers

PR #471, opened 2022-02-06
Update index.rst

mishasoundactive:
I am not a master in translations and in github, but at this stage I can offer this option.  Automatic translation was used with my logical, semantic edits/corrections.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -7,25 +7,25 @@
 
 .. _index:
 
-Mixxx |version| User Manual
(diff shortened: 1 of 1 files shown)


Holzhaus:
Thanks for the effort, but we're using Transfex for translations. Please have a look at https://www.transifex.com/mixxx-dj-software/mixxxdj-manual/ and click the 'Help Translate "Mixxx DJ Manual"' button in the top right corner.

PR #474, opened 2022-02-24
Traktor S3: Update Button Pads section for new sampler behavior

ywwg:
By default, both sides of the controller launch samplers 1-8


ywwg:
corresponds to https://github.com/mixxxdj/mixxx/pull/4676

Swiftb0y (approved):

Issue #473, opened 2022-02-12
Add info about Left / Center / Right buses


ronso0:
noticed in [No samplers sound with external mixer](https://mixxx.discourse.group/t/no-samplers-sound-with-external-mixer/24236)

PR #489, opened 2022-06-18
Preferences: explain sound I/O device settings

ronso0:
https://github.com/mixxxdj/manual/issues/473

https://deploy-preview-489--mixxx-manual.netlify.app/chapters/preferences/sound_hardware.html#output-and-input-devices


daschuer (approved):
Thank you.
Does anyone else like to have a look?

The CI failure is unrelated.

source/chapters/preferences/sound_hardware.rst
```diff
@@ -183,3 +183,56 @@ Other Sound Hardware options
...
+
+Inputs
+^^^^^^
+
+* **Vinyl Control 1-4**: Inputs to be used for :ref:`vinyl control and passthrough <vinyl-control>`.
+  Note that one input can be selected for multiple decks.
+
+* **Microphone 1-4**: ref:`Microphone <microphones>` inputs.
+
+* **Auxiliary 1-4**: Auxiliary inputs.
```
Swiftb0y, line 234: I think this needs further explanation or at least a `:ref:` to where these are explained.

source/chapters/preferences/sound_hardware.rst
```diff
@@ -183,3 +183,56 @@ Other Sound Hardware options
...
+------------------------
+
+In the usual, basic 'Digital DJ' setup (:ref:`setup-controller-and-external-card`
+or :ref:`setup-laptop-and-external-card`) all that needs to be configured are
+the Master and Headphones outputs. Though, there are many more output and
+input channels available for more complex setups, for example with external
+mixers, vinyl control or for broadcasting.
+
+All outputs are usually sent to a channel pair (stereo), though they can also be sent to one channel only (mono mixdown), for example if you're short on output channels, like when you need to use one stereo output for both master and
+headphones (see :ref:`setup-laptop-with-splitter`).
```
Swiftb0y, line 197: fix linebreak
```suggestion
All outputs are usually sent to a channel pair (stereo), though 
they can also be sent to one channel only (mono mixdown), for example
if you're short on output channels, like when you need to use one
stereo output for both master and headphones (see :ref:`setup-laptop-with-splitter`).
```

source/chapters/preferences/sound_hardware.rst
ronso0: You refer to Aux? so would you write a chapter? 😉

Swiftb0y: yikes, sorry I wasn't aware there wasn't any info on auxiliary inputs yet. Then can we leave a TODO instead?

ronso0: Sure. And a github issue would be helpful to collect basic info, as well as some hints/tweaks. Do you mind filung one?

Swiftb0y: #490

Be-ing:
I think this information would be more easily found if it was directly in the GUI as tooltips.

ronso0:
Yes, the I/O channel infos can also be used for the tooltips. The intro, general info and hints should be in the manual. Here we also link to related chapters, like broadcasting, mics and examole setups.

source/chapters/preferences/sound_hardware.rst
ronso0: done

source/chapters/preferences/sound_hardware.rst
ronso0: :+1: thanks!

ronso0:
Ready.
This is supossed to be the first step so we can link to the manual when the bus question comes up in the forums again.

Swiftb0y (approved):
Thanks LGTM

PR #491, opened 2022-06-21
Fix broken tables of hotcue_x controls

ronso0:
before:
https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-hotcue_X_enabled

now:
https://deploy-preview-491--mixxx-manual.netlify.app/chapters/appendix/mixxx_controls.html#control-[ChannelN]-hotcue_X_enabled

### DIFF
--- a/source/chapters/appendix/mixxx_controls.rst
+++ b/source/chapters/appendix/mixxx_controls.rst
@@ -1193,7 +1193,7 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
 
    Indicates if :term:`hotcue` slot X is set, active or empty.
 
-   :range: read-only
+   :range (read-only):
       ===== ===================================
       Value Meaning
       ===== ===================================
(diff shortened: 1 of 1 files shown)


ronso0:
This is just a quick fix which expands the first column. Any other idea how to fix this?

Holzhaus:
We could also make it a separate field, like "Writable: No".

Holzhaus (approved):
LGTM, thank you

ronso0:
> We could also make it a separate field, like "Writable: No".

a "field" like `range` or do you mean a separate column?

Holzhaus:
> > We could also make it a separate field, like "Writable: No".
> 
> a "field" like `range` or do you mean a separate column?

yes, a field like `range`. But I'm not sure if this takes too much space.

ronso0:
worth a try.
where are fields defined? didn't find it with grep range

PR #496, opened 2022-07-05
[DO NOT MERGE] Debug Scour failings on CI

Swiftb0y:
Many PRs currently don't pass CI because the scour pre-commit hook fails even though (according to the PR authors) it passes locally. This PR tries to debug that, **so don't merge it**.


Swiftb0y:
probably stale. i don't even remember what the original issue was about.

PR #502, opened 2022-08-21
add note about soundcard requirement

jevilalta:
Updating the SB3 documentation to address issue brought up in this thread:
https://mixxx.discourse.group/t/questions-about-ddj-sb3-and-decks-3-and-4/25539


ronso0:
Thanks for the quick fix!

PR #508, opened 2022-09-02
Redirect fix

daschuer:
This fixes the redirect to language with country specifier

### DIFF
--- a/build_html.sh
+++ b/build_html.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Parse languages and ensure that "en" is built last
 # This is necessary because we generate the redirect rules in the same loop
 # and want to use "en" as a fallback.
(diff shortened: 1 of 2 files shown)


daschuer:
Netlify was not smart enough for the underscore. This PR fixes the issue. @Swiftb0y can you have a look?

Swiftb0y:
I'm not sure how to test this to be honest.

daschuer:
We can only test post merge since it is a Netifly file.

build_html.sh
```diff
@@ -24,7 +24,7 @@ do
         printf '/:version/*    /:version/en/:splat    301\n' >> build/html/_redirects
     else
         printf '/:version/%s/* /:version/%s/404.html  404\n' "$lang" "$lang" >> build/html/_redirects
-        printf '/:version/*    /:version/%s/:splat    301 Language=%s\n' "$lang" "$lang" >> build/html/_redirects
+        printf '/:version/*    /:version/%s/:splat    301 Language=%s\n' "$lang" "${lang/_/-}" >> build/html/_redirects
```
Swiftb0y, line 27: isn't this the wrong way around. netflify languages contain a dash but our pages contain underscores?

daschuer: In the sphinx config we use now the underscores. This one replaces the underscores with dashes for Netlify so I think it should work.

Holzhaus:
> We can only test post merge since it is a Netlifly file.

This is not entirely true, you can just log in to netlify.app, create a new site and and then deploy to it. I did this when doing the initial migration to netlify. But if it's currently broken, the worst that can happen is that it's still broken. so we might as well merge.

Swiftb0y:
You're right this won't break it more than it already is, so we might as well merge.

PR #509, opened 2022-09-03
Language cleanup

daschuer:
This is a language clean up PR, which merges country specific translations in their base language and removes faulty and unmaintained translations.  
See single commits


Swiftb0y (approved):
Thanks. Local test looks good. Though some of the added languages have barely any strings translated.

daschuer:
I have not added any languages. The added files are a result from added pages and therefor it is quite natural that they are not yet translated.

Swiftb0y:
Ah yes, correct.

PR #512, opened 2022-09-04
sphinx_mixxx: declare safe for parallel reading and writing

daschuer:
This should silence the build warning:
```
WARNING: the sphinx_mixxx extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
WARNING: doing serial read
```

### DIFF
--- a/sphinx-mixxx/sphinx_mixxx/setup.py
+++ b/sphinx-mixxx/sphinx_mixxx/setup.py
@@ -23,4 +23,8 @@ def setup(app):
         "",
     )
 
-    return {"version": "0.1"}
+    return {
+        "version": "0.1",
+        "parallel_read_safe": False,
+        "parallel_write_safe": False,
+    }


Swiftb0y:
The PDF build fails with a strange error. This might be because of the parallel reading/writing. Not sure... Have you tried just explicitly specifying `parallel_read_safe`/`parallel_write_safe` as `False`? That should still silence the build error since it only complains about both not being specified.

daschuer:
I would rather hand this over to @Holzhaus instead of setting things I have no clue of. 
Hopefully he can confirm if the extension is usable for parallel builds or not.

Holzhaus:
I think parallel reading/writing may be a rpoblem because it builds a global list of Mixxx controls. But I'm not that deep into Sphinx' internals to be sure. I agree with @Swiftb0y that we can just set them to `False` to be safe. Then the warning should disappear without changing anything else (I think that sphinx just assumes `False` if that value is not set). The performance gains of setting this to `True` are probably neglible anyway.

daschuer:
Ok, I have now changed it to False.

Holzhaus (approved):
LGTM, thank you.

PR #520, opened 2022-10-01
Eq band bypass

ferranpujolcamins:
https://github.com/mixxxdj/mixxx/pull/10934


ronso0:
This is about a 2.4 feature so the manual update also needs to go to main, not 2.3
(though I'm a bit confused why there is a 2.4 and a main branch, https://mixxx.zulipchat.com/#narrow/stream/109176-documentation/topic/2.2E4.20branch.20of.20the.20manual)

PR #521, opened 2022-10-06
fix(ci): bump python runtime version

Swiftb0y:
attempt to fix netlify deploys such as 
https://app.netlify.com/sites/mixxx-manual/deploys/633e9ddf75531a000aa35ac3
(I have no idea what I'm doing).


Swiftb0y:
I guess bumping to python 3.8 worked. No idea why 3.7 broke.

Holzhaus:
LGTM. One thing that we could try is to set the runtime it to just `3` instead of `3.8`. Maybe that prevents breakage when they update Python to 3.9 or 3.10...

Holzhaus:
I think if you remove it entirely it will use python 2 (which has been EOL for years but I think the runners are debian-based...).

Swiftb0y:
flat out removing it works too :shrug:

Swiftb0y:
`3.*` works, but I'm not sure if that's because its actually supported or because a shells wildcard-expansion kicks in...

Holzhaus:
> `3.*` works, but I'm not sure if that's because its actually supported or because a shells wildcard-expansion kicks in...

The worst that can happen I that it breaks at some point and we have to intervene. Which is almost guaranteed if we hardcode 3.8.. Thanks for investigating.

Btw, I have no idea why it picks Python 2.7 when we don't pin a version, because [according to the docs 3.8 should be the default](https://github.com/netlify/build-image/blob/focal/included_software.md) and we are using focal (looked it up in the Netlify settings).

daschuer:
This is an issue due to changing from the deprecated Ubuntu Bionic Server to Ubuntu Focal. 
Thank you for fixing this. I will try the asterisk solution for our webside as well.

PR #525, opened 2022-12-23
Fix pre commit

JoergAtGithub:
Pre-Commit didn't work for me locally on Windows:

1.) When I execute pre-commit on my command line, a dialog pops up, asking for my GitLab credentials - I have no account at GitLab
2.) Python interpreter is not found (same fix as in https://github.com/mixxxdj/mixxx/pull/11004 )

PR #529, opened 2023-01-15
Pre commit fix2.3

JoergAtGithub:
Same as #525, but with target 2.3

Pre-Commit didn't work for me locally on Windows:

1.) When I execute pre-commit on my command line, a dialog pops up, asking for my GitLab credentials - I have no account at GitLab
2.) Python interpreter is not found (same fix as in https://github.com/mixxxdj/mixxx/pull/11004 )


JoergAtGithub:
I don't know why CI fails for target 2.3, but the identical change with target main ( in #525) passed.

daschuer:
It fails here, because the 2.4 version is already optimized:
See:
`git diff upstream/2.3 upstream/2.4 -- source/_static/controllers/hercules_djcontrol_mp3_e2_mapping.svg`
I think you can just apply the proposed patch. 
The link issue is fixed here: 
https://github.com/mixxxdj/manual/pull/532

daschuer (approved):
Thank you. LGTM

PR #532, opened 2023-01-30
Fix DDJ 400 links

daschuer:


source/conf.py
```diff
@@ -449,6 +449,7 @@
     "https://www.numark.com/images/product_downloads/MixtrackPro3-UserGuide-v1.1.pdf",
     "https://www.numark.com/product/*",
     "https://www.pioneerdj.com/en/support/software/controller/*",
+    "https://support.pioneerdj.com/hc/sections/*",
```
Swiftb0y, line 452: I guess pioneer doesn't like it all when they're site is scraped?

Issue #553, opened 2023-03-24
mixbot writes corrupt changelog.rst files, which let Sphinx / Build Check fail and inhibits generation of preview


JoergAtGithub:
How can we prevent, that mixbot can commit files, that break builds of HEAD?
https://github.com/mixxxdj/manual/actions/runs/4378433738/jobs/7665114625?pr=550

Swiftb0y:
Should be straightforward, seems to me like we're just missing an underscore? https://github.com/sphinx-doc/sphinx/issues/3921

JoergAtGithub:
Is it intended, that the Sphinx checks are only executed for Pull-Request:
https://github.com/mixxxdj/manual/blob/2.3/.github/workflows/sphinx.yml
This way nobody will see if a push breaks the buildability until a Pull-Request is based on this branch.

Swiftb0y:
`.github/workflows/build.yml` builds and deploys on each push iiuc.

JoergAtGithub:
Yes, it builds and deploys, but doesn't check the code using Sphinx. Therefore the CI runs of the branches pass, but all PRs based on them fail.

Swiftb0y:
I don't understand, shouldn't the build fail then? I don't quite understand the difference.

JoergAtGithub:
If I understand it correct these are different runners.
-sphinx.yml only checks the code
-build.xml builds the HTML output and deploys it

The failure is only detected by Sphinx, but not during the HTML generation - and Sphinx is not executed for pushes.

Swiftb0y:
but build.xml uses sphinx to build the html and pdf outputs. So why does it not fail while building?

JoergAtGithub:
@daschuer The 2.4 changelog broke again the manual builds:
`Warning: /home/runner/work/manual/manual/source/chapters/appendix/changelog.rst:10: WARNING: Duplicate explicit target name: "#11460".`

PR #556, opened 2023-04-03
Fix linkcheck

JoergAtGithub:
This PR:
-Changes a link to the dead DJTechTools forum to a comment with a reference to the source (the forum is not archived by webarchive)
-Enables the execution of build-check and link-check for push events as well. To detect issues like #553


daschuer (approved):
Thank you.

PR #565, opened 2023-06-03
fix(changelog): workaround changelog entries causing build errors

### LINKED ISSUE #553: mixbot writes corrupt changelog.rst files, which let Sphinx / Build Check fail and inhibits generation of preview

Swiftb0y:
Fixes #553

Somewhat dirty workaround but good enough I think. wdyt?

### DIFF
--- a/tools/update_changelog.py
+++ b/tools/update_changelog.py
@@ -42,7 +42,13 @@ def changelog_to_rst(changelog):
         flags=re.MULTILINE,
     )
 
-    return TEMPLATE.lstrip().format(content=m2r2.convert(changelog))
+    changelog = TEMPLATE.lstrip().format(content=m2r2.convert(changelog))
+
+    # m2r2 produces links with postfix _. We need anonymous references instead
+    # See https://github.com/mixxxdj/manual/issues/553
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
The underscore suffix looks wrong, but Sphinx currently fails because we have the link name multiple times:
`Warning: /home/runner/work/manual/manual/source/chapters/appendix/changelog.rst:10: WARNING: Duplicate explicit target name: "#11460"`
https://github.com/mixxxdj/manual/blob/66b9ebee8eeaf42d4dc3d800bab6b925865dafcd/source/chapters/appendix/changelog.rst?plain=1#L550-L556

Swiftb0y:
Yeah, thats because each of those links creates a target and there can't be multiple of them. This duplicates the underscore at the end, which will result in anonymous references, which won't create a target and thus the link text is irrelevant. The build failed previously, because the base commit was outdated.

Swiftb0y:
all green. finally

JoergAtGithub:
Thanks for the explanation. This makes sense to me! Thank you!

Swiftb0y:
Thank you.

PR #571, opened 2023-07-09
Native Instruments Traktor Z1: add controller to manual

fasterthanwhatever:
https://github.com/mixxxdj/mixxx/pull/11731

### DIFF
--- a/source/hardware/controllers/native_instruments_traktor_kontrol_z1.rst
+++ b/source/hardware/controllers/native_instruments_traktor_kontrol_z1.rst
@@ -1,0 +1,11 @@
+Native Instruments Traktor Kontrol Z1
+=====================================
+
+-  `Manufacturer’s product page <https://www.native-instruments.com/en/products/traktor/dj-controllers/traktor-kontrol-z1/>`__
+
+.. versionadded:: 2.4.0
+
+.. note::
+   Unfortunately a detailed description of this controller mapping is still missing.
+   If you own this controller, please consider
+   `contributing one <https://github.com/mixxxdj/mixxx/wiki/Contributing-Mappings#user-content-documenting-the-mapping>`__.


source/hardware/controllers/native_instruments_traktor_kontrol_z1.rst
```diff
@@ -0,0 +1,11 @@
...
+=====================================
+
+-  `Manufacturer’s product page <https://www.native-instruments.com/en/products/traktor/dj-controllers/traktor-kontrol-z1/>`__
+
+.. versionadded:: 2.4.0
+
+.. note::
+   Unfortunately a detailed description of this controller mapping is still missing.
+   If you own this controller, please consider
+   `contributing one <https://github.com/mixxxdj/mixxx/wiki/Contributing-Mappings#user-content-documenting-the-mapping>`__.
```
JoergAtGithub, line 11: The manual page for the Traktor Z1 needs to be written of cause, as described in the linked document.

JoergAtGithub (changes requested):

source/hardware/controllers/native_instruments_traktor_kontrol_z1.rst
fasterthanwhatever: if, i'm 100% honest. i found the mapping file in the www (https://github.com/xeruf/mixxx-controller-mappings). writing comprehensive documentation will be beyond my capabilities. sorry.

ronso0: I think this should also be mentioned in https://github.com/mixxxdj/mixxx/pull/11731?

Correct me if I'm wrong:
this is a basic two-channel mixer (not the F1 or X1), there are no layers (Shift or other modifiers/toggles) as far as I can tell, so documenting those basic functions should be pretty straight-forward. Even if you just mention that "all controls act as labeled", plus some words about the Fx knob.
Easy, no?
I think this would help other users to decide whether to try Mixxx (if they own the Z1) or pick the Z1 for use with Mixxx.

Fact is: probability that some other _user_ will open a PR to document the Z1 is rather low...

JoergAtGithub:
Superseded by #594

PR #573, opened 2023-07-12
remove slightly insensitive genre filtering example.

ywwg:
There are racial overtones to the term "gangsta" that I would prefer we just avoid completely.


Swiftb0y (approved):
country hip-hop doesn't make much sense but neither does specifying the genre twice IMO, so LGTM.

Holzhaus:
How about distinguishing rap and instrumental hip hop?

PR #576, opened 2023-08-06
Denon MC6000MK2: Update master level mapping

uklotzde:
Follow-up


uklotzde:
Performing the link check on every PR is pointless and confusing. PRs fail for unrelated reasons.

Instead the link check should be performed periodically. Then someone could independently fix those issues.

Swiftb0y (approved):

PR #584, opened 2023-10-06
Document the new "OR" operator

fwcd:
As a sibling PR to https://github.com/mixxxdj/mixxx/pull/12061, this documents the new "OR" operator.

https://deploy-preview-584--mixxx-manual.netlify.app/chapters/library#finding-tracks-search

### DIFF
--- a/source/chapters/library.rst
+++ b/source/chapters/library.rst
@@ -440,7 +440,6 @@ Mixxx supports the following filters:
 
        ~key:c#m
 
-    You can combine operators but there's no way to do an “OR” search right now.
     The following example lists all tracks by “Danger” over 3 minutes long that
     are rated 4 or 5.
 
@@ -474,6 +473,28 @@ Mixxx supports the following filters:
   .. note::
      You can put a space after the colon but currently there must be no space
      between the operator and the number.
+
+In addition to combining filters directly (effectively an "AND"), you can also combine queries using the "OR" operator, spelled ``|``.
+
+Examples
+  ::
+
+    genre:house | genre:techno
+
+  The "OR" operator has a lower precedence than regular filters, so you can create larger queries too. The following query, for example, would match all tracks that are Pop or 5-star Hip-Hop:
+
+  ::
+
+    genre:pop | rating:5 genre:hip-hop
+
+  Similar to popular search engines, the "OR" operator also supports the alternative spelling ``OR``, so the previous query could equivalently be written as:
+
+  ::
+
+    genre:pop OR rating:5 genre:hip-hop
+
+.. note::
+  The spelled-out form is case-sensitive: While uppercase ``OR`` refers to the operator, lowercase ``or`` refers to the literal string ``or``. Using the quoted form ``"OR"`` will also search for the literal string.
 
 .. _library-previewing-tracks:
 


source/chapters/library.rst
```diff
@@ -475,6 +474,28 @@ Mixxx supports the following filters:
...
      between the operator and the number.
 
+In addition to combining filters directly (effectively an "AND"), you can also combine queries using the "OR" operator, spelled ``|``.
+
+Examples
+  ::
+  
+    genre:house | genre:techno
+
+  The "OR" operator has a lower precedence than regular filters, so you can create larger queries too. The following query, for example, would match all tracks that are 5-star Hip-Hop or Pop:
```
ronso0, line 484: I'm not sure if that means
[5-star Hip-hop] | Pop
or
5-star [Hip-hop | Pop]
Can we rephrase it
```suggestion
  The "OR" operator has a lower precedence than regular filters, so you can create larger queries too. The following query, for example, would match all 5-star tracks of either Hip-Hop or Pop genre:
```

source/chapters/library.rst
```diff
@@ -475,6 +474,28 @@ Mixxx supports the following filters:
...
+
+  Similar to popular search engines, the "OR" operator also supports the alternative spelling ``OR``, so the previous query could equivalently be written as:
+
+  ::
+
+    genre:hip-hop rating:>=4 OR genre:pop
+
+.. note::
+  The spelled-out "OR" operator is case-sensitive, i.e. searching for ``or`` will search for the literal string ``or``. Using the quoted form ``"OR"`` will also search literally.
+
```
ronso0, line 498: I feel that, in this place, this important section does not get the attention it deserves.
What do you think about moving it up, below the first paragraph?

> ### Using search operators
> 
>Search operators allow you to form more complex search queries. They allow you to limit certain search terms to particular properties of your tracks.

<--

> Mixxx supports the following filters:

ronso0: The failing link check is unrelated (we need to merge 2.3 into 2.4, 2.4 into main) and pre-commit only found a trailing whitespace.

source/chapters/library.rst
fwcd: It it supposed to mean `[5-star Hip-hop] | Pop` (since the OR binds more loosely than AND) and I agree, it does feel a bit ambiguous. Not sure how to phrase it better though, I feel like the suggestion leans itself more towards the second interpretation.

Note that the second one cannot actually be expressed with the current machinery since we don't have any parentheses yet (if I haven't missed something).

fwcd: Perhaps simply flipping the operands would make it more clear:

```suggestion
  The "OR" operator has a lower precedence than regular filters, so you can create larger queries too. The following query, for example, would match all tracks that are Pop or 5-star Hip-Hop:
```

source/chapters/library.rst
fwcd: My idea was to first introduce the "OR" operator itself, then the spelling and finally discuss this rather unusual (but not inconceivable) edge case/gotcha. Note that strings only containing "OR" are unaffected, so I would expect such queries to be pretty uncommon. Users searching in lowercase would never run into this anyway.

Also I think it would be nice to keep everything about the "OR" operator in a single section.

Could you elaborate on why this note deserves more attention?

ronso0: > Also I think it would be nice to keep everything about the "OR" operator in a single section.

Of course. I think this was just a misunderstanding, by 'section' I mean the entire new OR section.
Anyway, the placement is fine for now, maybe it's just that I find the current stylesheet a bit inconclusive, so it's a general issue IMHO. With the bullet points it looks like OR belongs **Special filtering**. 
Maybe it would help if OR gets a headline **Combine filter with OR**.
Later on we may move some AND examples from **Special filtering** to **Combine filters** where we clarify the default AND and explicit OR. But that's out of scope here.

ronso0:
Thank you very much for this PR (and the OR PR itself!).
pre-commit is still complaining about a trailing whitespace.

Please remove that, then squash all commits and force-push?

source/chapters/library.rst
fwcd: Ah sorry, yeah, that was a misunderstanding on my part. That sounds good!

ronso0:
Thank you!

PR #590, opened 2023-10-11
Library: document new text filter operator '=' (find exact matches)

ronso0:
document https://github.com/mixxxdj/mixxx/pull/12063

https://deploy-preview-590--mixxx-manual.netlify.app/chapters/library#using-search-operators

### DIFF
--- a/source/chapters/library.rst
+++ b/source/chapters/library.rst
@@ -413,6 +413,19 @@ Mixxx supports the following filters:
      It doesn't matter if you have space between the colon and the argument
      or not. Quotes must be used for multi-word text arguments.
 
+  You can use **=** to find exact matches. It reveals only tracks where the entire property
+  text equals the search term. As without **=**, you must use quotes for more than one word.
+  The first of the examples below will find only tracks where the title is
+  "wow" (case-insensitive like the default search), i.e. not "wowy", "oh wow!"
+  or "wow (Hej Remix)".
+
+  Examples
+    ::
+
+      title:=wow
+      title:= "Track 1"
+      artist:="DJ Flop"
+
 * **Numeric filtering**: bitrate, bpm, played, rating, track, year
 
   Examples


source/chapters/library.rst
```diff
@@ -413,6 +413,19 @@ Mixxx supports the following filters:
      It doesn't matter if you have space between the colon and the argument
      or not. Quotes must be used for multi-word text arguments.
 
+  You can use the **=** operator to find only tracks where the entire property
+  text equals the search term. You can also enclose multiple terms in quotes.
+  The first of the examples below will find only tracks where the title is
+  "wow" (case-insensitive like the default search), i.e. not "wow!", "wowy"
+  or "wow (Hej Remix)".
```
Swiftb0y, line 420: I think there should also be an example with a prefix, eg: 
```suggestion
  "wow" (case-insensitive like the default search), i.e. not "wowy", "oh wow!"
  or "wow (Hej Remix)".
```

source/chapters/library.rst
```diff
@@ -413,6 +413,19 @@ Mixxx supports the following filters:
      It doesn't matter if you have space between the colon and the argument
      or not. Quotes must be used for multi-word text arguments.
 
+  You can use the **=** operator to find only tracks where the entire property
+  text equals the search term. You can also enclose multiple terms in quotes.
```
Swiftb0y, line 417: I think this is clearer.
```suggestion
  You can use the **=** operator to find only tracks where the entire property
  text equals the search term. As without **=**, you must use quotes for more than one word.
```

Swiftb0y (approved):
LGTM. thank you for cleaning up the loose ends. Any1 else wanting to review?

daschuer:
I am missing something like the PR headline in the text.

ronso0:
What exactly do you have in mind?

IMHO this is clarified with
> You can use the = operator to find only tracks where the entire property text equals the search term.

Note that this is just an extension, not an entirely new feature which would deserve more emphasis.

source/chapters/library.rst
```diff
@@ -413,6 +413,19 @@ Mixxx supports the following filters:
      It doesn't matter if you have space between the colon and the argument
      or not. Quotes must be used for multi-word text arguments.
 
+  You can use the **=** operator to find only tracks where the entire property
```
daschuer, line 416:
```suggestion
  You can use **=** for finding exact matches. It reveals only tracks where the entire property
```

ronso0:
Alright, done.

daschuer (approved):
Thank you.

PR #596, opened 2023-12-09
Replace another outdated broken link to manufacturer page for A&H Kone K2

JoergAtGithub:


ronso0:
Thank you, now all K2 links are fixed ;)

PR #611, opened 2024-02-05
fix `beatjump` description and beatsize range

ronso0:


source/chapters/appendix/mixxx_controls.rst
```diff
@@ -510,9 +512,9 @@ Any control listed above for :mixxx:cogroupref:`[ChannelN]` will work for a samp
                    [PreviewDeckN],beatjump
                    [SamplerN],beatjump
 
-   Jump forward by :mixxx:coref:`beatjump_size <[ChannelN],beatjump_size>` (positive) or backward by :mixxx:coref:`beatjump_size <[ChannelN],beatjump_size>` (negative). If a loop is active, the loop is moved by :mixxx:coref:`beatjump_size <[ChannelN],beatjump_size>`.
+   Jump forward (positive) or backward (negative) by N beats. If a loop is active, the loop is moved by X beats.
```
Holzhaus, line 515: Might make sense to split these long lines and put each sentence on a separate lines if you touch them anyway. Your decision.

Holzhaus (approved):
LGTM.

source/chapters/appendix/mixxx_controls.rst
ronso0: I'm not sure whether the line breaks would really be helpful here. If applied consistently for all controls, the page would grow enormously when being viewed on mobile screens.

Holzhaus: Oh, I meant regular line break to make diff nicer to read. I thought this was not make da difference in the resulting html unless you put a blank line in between. Is that incorrect?

PR #630, opened 2024-03-22
Looping: mention seconds fallback in case track has no beats

ronso0:
Documentation for new seconds fallback if track has no beats.
https://github.com/mixxxdj/mixxx/pull/12961


Swiftb0y (approved):
LGTM. Thank you.

PR #631, opened 2024-03-25
Manuel fr 001

Eve00000:


daschuer:
Thank you for the PR. You have probably started from the main branch that targets our 2.5 release, was this intended?

Eve00000:
> Thank you for the PR. You have probably started from the main branch that targets our 2.5 release, was this intended?

Antoine told me to make a rp so we could work on the french translation together. But I saw there were much files changed so something must be wrong. So the problem is that I started with the 2.5 ? How can that be seen?
I just translated one line as a test.
Can I proceed like this or ..... please tell me (as long as it isn't transiflex)

daschuer:
Do you have git installed? It is "Git for Windows" if you are on windows. 
Than clone your private repository via copy the following line by line to "Git Bash":  

```
cd Documents
git clone https://github.com/Eve00000/manual.git
cd manual 
git remote add upstream https://github.com/mixxxdj/manual.git
git fetch upstream
git checkout Manuel-Fr-001
git rebase --onto=upstream/2.4 a95da34b00d1eb915b1f0077e65fa4692ff384d0
git push
```  

You will after that find the po files in 
C:\Users\Eve00000\Documents\manual\locale\fr\LC_MESSAGES\....

After editing you can start "Git Bash" again and: 
```
cd Documents
cd manual 
git commit -a -m"Manuel Français" 
git push
```  
Done

PR #678, opened 2024-06-19
Move track commands

nlw0:
Documenting commands from https://github.com/mixxxdj/mixxx/pull/13098#issuecomment-2176861227

Preview:
https://deploy-preview-678--mixxx-manual.netlify.app/chapters/appendix/mixxx_controls#control-[Library]-MoveTrackUp

### DIFF
--- a/source/chapters/appendix/mixxx_controls.rst
+++ b/source/chapters/appendix/mixxx_controls.rst
@@ -3204,6 +3204,16 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
    .. versionadded:: 2.5.0
 
 
+.. mixxx:control:: [Library],MoveTrack
+
+   Move track selection up or down by one row in the playlist. Equivalent to pressing the :kbd:`Alt` + :kbd:`Up` or :kbd:`Alt` + :kbd:`Down` key on the keyboard. This is applicable only in unlocked playlists.
+
+   :range: Binary
+   :feedback: Currently selected tracks are moved
+
+   .. versionadded:: 2.5.0
+
+
 .. mixxx:control:: [Library],ScrollUp
 
    Equivalent to pressing the :kbd:`PageUp` key on the keyboard
--- a/source/chapters/appendix/mixxx_controls.rst
+++ b/source/chapters/appendix/mixxx_controls.rst
@@ -3174,6 +3174,26 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
    .. versionadded:: 2.1.0
 
 
+.. mixxx:control:: [Library],MoveTrackUp
+
+   Equivalent to pressing the :kbd:`Alt` + :kbd:`Up` key on the keyboard
+
+   :range: Binary
+   :feedback: Move selected track up
+
+   .. versionadded:: 2.5.0
+
+
+.. mixxx:control:: [Library],MoveTrackDown
+
+   Equivalent to pressing the :kbd:`Alt` + :kbd:`Down` key on the keyboard
+
+   :range: Binary
+   :feedback: Move selected track down
+
+   .. versionadded:: 2.5.0
+
+
 .. mixxx:control:: [Library],MoveVertical
 
    Move the specified number of locations up or down. Intended to be mapped to an encoder knob.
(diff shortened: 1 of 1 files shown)


source/chapters/appendix/mixxx_controls.rst
```diff
@@ -3174,6 +3174,26 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
...
+.. mixxx:control:: [Library],MoveTrackDown
+
+   Equivalent to pressing the :kbd:`Alt` + :kbd:`Down` key on the keyboard
+
+   :range: Binary
+   :feedback: Move selected track down
+
+   .. versionadded:: 2.5.0
+
+
```
ronso0, line 3196: Note that the controls seem to be ordered alphabetically, but actually it's better to have them grouped by 'function'/effect. Please move the block down so `MoveVertical` is below `MoveUp` again.

source/chapters/appendix/mixxx_controls.rst
```diff
@@ -3174,6 +3174,26 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
    .. versionadded:: 2.1.0
 
 
+.. mixxx:control:: [Library],MoveTrackUp
+
+   Equivalent to pressing the :kbd:`Alt` + :kbd:`Up` key on the keyboard
```
ronso0, line 3179: I'm not sure everyone browsing these controls is aware of the new Alt move feature.
I suggest to add more context here in order to make this feature more discoverable.
For example:
> Move the track selection up by one row. This is applicable only in (unlocked) Playlists.

Same for down and MoveTrack.

source/chapters/appendix/mixxx_controls.rst
```diff
@@ -3174,6 +3174,26 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
...
+
+   :range: Binary
+   :feedback: Move selected track up
```
ronso0, line 3182: or put it here (but maybe the explanation is to much for the 'feedback' hint):
> Track selection is moved up by one row

What do you think?

ronso0 (changes requested):
Thank you for this!

Some notes

source/chapters/appendix/mixxx_controls.rst
nlw0: Oh sure, I didn''t realize it was related

source/chapters/appendix/mixxx_controls.rst
nlw0: I reworded a bit both texts, what do you think?

source/chapters/appendix/mixxx_controls.rst
nlw0: I reworded it so it's a bit more like other entries

ronso0:
Thanks, looks good.

Did you accidentally remove `MoveTrack`?

nlw0:
> Thanks, looks good.
> 
> Did you accidentally remove `MoveTrack`?

I had not added it at first, it's there now.

source/chapters/appendix/mixxx_controls.rst
```diff
@@ -3184,6 +3184,36 @@ Note that :mixxx:coref:`[Library],MoveUp` and other Move and Scroll controls emu
...
+   :feedback: Currently selected tracks are moved
+
+   .. versionadded:: 2.5.0
+
+
+.. mixxx:control:: [Library],MoveTrack
+
+   Move track selection up or down by one row in the playlist. Equivalent to pressing the :kbd:`Alt` + :kbd:`Up` or :kbd:`Alt` + :kbd:`Down` key on the keyboard. This is applicable only in unlocked playlists.
+
+   :range: Binary
```
ronso0, line 3211: This should actually be [`integer` (move by n rows)](https://github.com/mixxxdj/mixxx/blob/4c11d9b538cf8d27acdef6665ce1b54658b7690d/src/library/librarycontrol.cpp#L824-L836), but I just noticed that this optional factor is not considered in [`WLibraryTableView::moveCursor()`](https://github.com/mixxxdj/mixxx/blob/a48562ad96021bd9571c8fbc8d331291bc30c1fc/src/widget/wlibrarytableview.cpp#L265).

I've just  filed https://github.com/mixxxdj/mixxx/issues/13398

ronso0:
@nlw0 Do you think you can finish this?
This is the last TODO is `:range: integer` for `[Library],MoveTrack`

nlw0:
I hope I can do it this weekend.

ronso0:
@nlw0 
> I hope I can do it this weekend.

Or next ; )

PR #682, opened 2024-07-13
Korg Kaoss DJ - macOS Monterey note

larromba:
Hello,

I noticed the Korg Kaoss DJ was no longer working on macOS. After a bit of research, I found that it wasn't Mixxx, but that it's only supported by Korg until macOS Monterey. Please search for `KAOSS DJ` [here](https://www.korg.com/download/global/support/os/pdf/mac_compatibilitychart_KORG_en.pdf)

The issue was raised here:
https://github.com/mixxxdj/mixxx/issues/13439 

Thanks


Swiftb0y (approved):
Thank you. LGTM. Also added to the wiki.

PR #690, opened 2024-08-18
Denon MC7000: Document new jogwheel settings

fwcd:
This is a sibling PR to https://github.com/mixxxdj/mixxx/pull/4684 that documents the new mapping settings, specifically jogwheel acceleration.


JoergAtGithub:
The formatting looks broken: 
grafik

fwcd:
Oh that's interesting, locally the math rendered fine for me:

image

Is this something that has to be configured for the deployment?

JoergAtGithub:
My screenshot is from the Netlify Preview.

fwcd:
Looks like something's wrong with the way MathJax is loaded:

image

I found a forum post that discusses this and we might be able to get this to work by updating the content security policy to allow loading the MathJax script:

https://answers.netlify.com/t/enabling-mathjax-without-using-cms/6873/4

We ran into this now because the Mixxx manual didn't seem to use any math yet, but if we need to add a formula somewhere in the future this would probably be useful to have anyway.

Swiftb0y:
I agree. We should also look to whether its feasible to render the formula to SVG at build-time rather than relying on client-side JS to do it since it results in a suboptimal user experience (slower loadtime, likely shifting page content as placeholders get expanded to their content, etc).

JoergAtGithub:
Maybe it's only an issue of the Netlify Preview?

fwcd:
> We should also look to whether its feasible to render the formula to SVG at build-time rather than relying on client-side JS to do it since it results in a suboptimal user experience

MathJax is pretty standard across the ecosystem and from what I understand client-side rendering is a bit better for accessibility, but you're right that things shifting around is generally less nice. Maybe MathML would be an option too, but I'm not sure if Sphinx supports that.

I think configuring the Netfliy CSP to support MathJax should be pretty easy, therefore I would lean towards doing that for now and moving SVG prerendering etc. into a separate PR.

fwcd:
Okay, so enabling SVG rendering was actually a lot easier than I thought, configuring Sphinx to use `sphinx.ext.imgmath` with SVG was enough 😄 Though I am still a bit torn. SVG rendering might be the easiest, least-invasive option for now, but I still think we should explore MathJax/MathML rendering in the future.

Advantages of SVG rendering:

- Faster rendering on the client-side
- Better compatibility with different (especially legacy) browsers
- Doesn't require changing the Content-Security-Policy

Disadvantages:

- Slower build times
- Worse accessibility
- Text selection doesn't work in formulas
- Text is less crisp due to being prerendered

fwcd:
Hm, looks like we'll still have to change the CSP, even for SVGs:

image

fwcd:
I suspect the issue is that Netlify's build environment is somehow missing `dvisvgm`... maybe MathJax is the way to go after all.

fwcd:
After updating the Content-Security-Policy to allow MathJax, the math now renders properly on Netlify. This uses the default client-side MathJax renderer for now, we can discuss using alternative renderers in future PRs including potential tradeoffs, this seems to be too large to tackle here.

I have also squashed/clean up my commits, so it should be ready to re-review.

JoergAtGithub:
LGTM! Thank you!

PR #691, opened 2024-08-21
Fix: Remove wrong shortcut reference from library.rst

cr7pt0gr4ph7:
The current documentation is outdated: As of the Mixxx 2.5 (and probably earlier), the `Esc` shortcut exits the search box, but does not clear its contents.


ronso0:
You're right, this should already be fixed for 2.4
I changed the base to 2.4, which created a huge diff, so I went ahead and merged you commit manually.

Thank you!

PR #716, opened 2025-01-15
Fix/ci upgrade python

Swiftb0y:
Usual maintenance as `3.7` got dropped in the CI runners/setup. Lets see.


daschuer:
Can we also pin the Ubuntu version, to prevent us from such an unstable regression in future?

Swiftb0y:
We can... but I'm unsure if there is an advantage. Sure CI won't break silently but upgrading (which we'll have to do inevitably at some point) will be harder because because so much changes at once...

source/conf.py
```diff
@@ -409,7 +409,7 @@
 # The format is a list of tuples containing the path and title.
 # epub_pre_files = []
 
-# HTML files shat should be inserted after the pages created by sphinx.
+# HTML files that should be inserted after the pages created by sphinx.
```
Swiftb0y, line 412: favorite typo so far :rofl:

Swiftb0y (commented):
ready for review

daschuer:
> We can... but I'm unsure if there is an advantage. Sure CI won't break silently but upgrading (which we'll have to do inevitably at some point) will be harder because because so much changes at once.

ubuntu-latest is no rolling release. Is just points to one of the workflow runners. When GitHub decide to relink it, we have the same bunch of changes as if we do the change manually. The difference is we can control the date and it will not happen during a release process or at some other bad timing. 

IMHO pinning the python version does make no sense if we do not pin the runner that has that version. I don't see any disadvantages.

Swiftb0y:
@Holzhaus, thoughts?

Holzhaus (approved):

PR #604, opened 2024-01-03
Updated Hercules Inpulse 300 manual to match #12503

BoredGuy1:
This pull request updates the wiki so that it is in line with [#12503](https://github.com/mixxxdj/mixxx/pull/12503).

PR #723, opened 2025-02-06
Hercules inpulse 300

acolombier:
Fix for #604 on which I missed my forced push, leading to the PR to close, and my write access to be revoked :sweat:


acolombier:
Merging directly as this is a pristine copy of #604, and blocking the mapping merging

PR #726, opened 2025-02-20
Pitch slider is no longer opposite to turntable/CDJ behaviour

bigpresh:
The manual states that:
> The slider allows you to change the speed of the song, by default
> up to 10% from the tracks original tempos. The speed will increase
> as you move the slider up, opposite to the behavior found on
> DJ turntables and CDJ

This appears to no longer be the case by default - the "Down increases speed" option under Decks in Preferences appears to be the default setting (and the default range appears to be 8% not 10% as previously documented).

Certainly my Mixxx 2.5.0 wearing the LateNight skin when reset to default preferences (moving `~/.mixxx` aside temporarily) has the "Down increases speed" option selected (and the range set to 8%, not 10% as the manual suggests)

I couldn't find an entry in the changelog to say if/when this changed.

### DIFF
--- a/source/chapters/user_interface.rst
+++ b/source/chapters/user_interface.rst
@@ -621,10 +621,13 @@ control rate changes also from your computer's keyboard, see the chapter
   :ref:`sync-lock` for more information.
 
 **Pitch/Rate slider**
-  The slider allows you to change the speed of the song, by default up to 10%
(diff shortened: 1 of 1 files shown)


daschuer:
I can confirm. The issue issue is that the default setting with the default button in the preference is of another opinion. This seems to be a bug. I will fix it as well.

PR #728, opened 2025-02-26
Add M-Vave SMK-25 II manual page

SamWhited:
This adds a manual page for the M-Vave SMK-25 II. I confirm that I have signed the CLA.

The corresponding Mixxx PR is here: https://github.com/mixxxdj/mixxx/pull/14412

Preview: https://deploy-preview-728--mixxx-manual.netlify.app/hardware/controllers/mvave_smk-25-ii


.gitignore
```diff
@@ -6,3 +6,4 @@
 build
 source/locale/pot/.doctrees
 /source/locale/en_US
+sphinx-mixxx/sphinx_mixxx.egg-info/
```
ronso0, line 9: Never seen this file, where does it come from?
I suspect this appears due to some custom cfg? If so, it belongs into your user ignore list.

SamWhited: It gets spit out by Python sometimes when you do `make html`. It's unclear to me why as I thought eggs were long deprecated, but it's sometimes appearing for me and sometimes not when building in a local venv.

SamWhited: Actually, I can't get this to happen now. Maybe I manually ran sphinx the first time or something? Unclear. Egg files definitely shouldn't be committed if they *are* generated somehow, but it doesn't appear to be a regular problem. I can drop the commit if you prefer.

ronso0: Hmm, the ignore section is `# Generated during a regular build` and I did a lot of 'regular builds' (html, pdf has been a while) so I prefer to remove it.
Let us know when it's bee created again.

SamWhited: Dropped the commit and squashed everything else for good measure. Thanks for the feedback!

SamWhited: I realized right after dropping this that I know exactly how to regenerate it; if you are installing sphinx itself (and other requirements) in a virtual env, eg. via `pip install requirements.txt`, this file will always be generated. If you're using a system wide installation of the theme and sphinx this isn't necessary, but for systems like mine that don't have one or more of these dependencies packaged (Arch is missing the SVG converter stuff) it's necessary to install them yourself using pip. Happy to leave it as is, or pull the commit back in. Whatever works best for you.

ronso0: I can confirm this: egg has been created by
`pip3 install -r requirements.txt --upgrade`

@Holzhaus what's your opinion on this?
Add to Mixxx' or personal ignore list?

SamWhited: In the mean time, is there anything else needed for this specific PR since this commit is unrelated and can be followed up on afterwards?

source/hardware/controllers/mvave_smk-25-ii.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. csv-table::
+   :header: "Action", "Name", "Function"
+   :widths: 15 10 70
+
+   "Play", ":hwlabel:`⏵`", "Play/pause playback"
+   "Pause", ":hwlabel:`⏸`", "If track is playing: stops the track and resets position to the main cue point
+
+   If playback is stopped: sets the main cue point"
+   "Pause", ":hwlabel:`⏸` (hold)", "Play the track from main cue point, release to stop playback and return to the main cue point. Playback must be initially stopped on the main cue point."
```
ronso0, line 73: I think this is behavior defined by "Cue mode" in the deck preferences?
If yes, let's mention that and link to the cue modes description like this
:ref:\`cue mode <interface-cue-modes>\`

source/hardware/controllers/mvave_smk-25-ii.rst
```diff
@@ -0,0 +1,83 @@
...
+   "Record", ":hwlabel:`⏺`", "Toggle recording the mix"
+   "Rewind", ":hwlabel:`⏪`", "Beatjump backwards"
+   "Fastforward", ":hwlabel:`⏩`", "Beatjump forwards"
+   "Deck 1",  ":hwlabel:`◀`", "Select deck 1"
+   "Deck 2",  ":hwlabel:`▶`", "Select deck 2"
+   "Reloop",  ":hwlabel:`UNDO`", "Activate/deactivate the previously active loop"
+
+.. [#] Bluetooth is not as reliable as a wired connection for live
+   performances, and may result in less smooth fader operation and higher
+   latencies.
```
ronso0, line 83: I'd move this to the top Note instead

.gitignore
SamWhited: This PR is green now so I created a follow up for the .gitignore issue in #736 so that it doesn't get forgotten about and we don't have to block this PR on an unrelated change.

SamWhited:
Looks like CI passed that time!

SamWhited:
Oops, the button labels must be something about different systems and the font changing. Let's just remove the images if this is the case, making and editing them is *very* painful and they're not really necessary, just nice looking. I don't really know enough about SVG to fix the little issues like that and don't want to spend any more time in Inkscape. I'll prepare a new patch removing them later today.

ronso0:
Please keep the images!
I'll take a look. Labels should be easy to fix, simplest would rendering the texts to paths.

SamWhited:
Will do, I'll hold off for now. I did ask on fedi if it's possible to automatically convert all text to paths when exporting for the web, I just really don't want to do it by hand. A few other comments:

> increase image size. currently they are width="84.361mm" height="50.167mm", let's use pixels and make it like 150% of the size used when embedded?

Does this matter in any way to the final output? As far as I know it just gets scaled up or down as needed and the base size doesn't matter at all (except maybe if it's really tiny and that becomes the default size on the web)? I couldn't figure out how to get the rendered size in the manual, if I view the "computed" properties on firefox it just lists size as "100%".

> remove blue bg from BT button?

This is just always the color of this button (it gets brighter when connected, dimmer when using USB-C or before bluetooth connects).

> does the display always show "Pr1"? If not let's remove it

This is what it shows by default, maybe I should have made it say "Mixxx!" or something (although in the original a 7-segment display font was used and there isn't actually room for that on the real display, but we could fudge it for the sake of it just being a fun illustration)!

ronso0:
> Does this matter in any way to the final output?

I don't think it matters for the manual page itself, but it apparently matters when the image is opened in a new tab. For me it's ~6x4 cm.

ronso0:
> This is just always the color of this button (it gets brighter when connected, dimmer when using USB-C or before bluetooth connects).

Ah okay. I just think that it stands out too much and makes it a bit unclear where the dashed "Layer Select" line is pointing to.
Maybe make the line go around the controls it crosses? Or just make it solid? Or add a white outline?

SamWhited:
Gotcha, thanks. Someone on fedi told me how to export and make sure all text shows up the same everywhere, I'll do that really quick and push up the changes, can you tell me if they look better on your end? I'll have the export increase the width at the same time

ronso0:
> This is what it shows by default, maybe I should have made it say "Mixxx!" or something ... we could fudge it for the sake of it just being a fun illustration)!

Hehe!

if you can write to it, maybe show the current deck number?

SamWhited:
> if you can write to it, maybe show the current deck number?

Sorry, I meant just in the image, as far as I can tell there's no way to write to it

SamWhited:
Applied the patch suggested by pre-commit, but now it seems to be suggesting another one? Unclear what's going on here.

This version has been exported with the "--export-width" option set to something larger and with the text-to-path option enabled, so hopefully it's better. I don't really want to have to go back into Inkscape to make actual changes though, this sort of thing isn't my area and it's *very* tedious.

source/hardware/controllers/mvave_smk-25-ii.rst
```diff
@@ -0,0 +1,81 @@
...
+distinct layers for playing samples and for controlling the deck.
+On both layers, the touch strips and rotary encoders are mapped as follows:
+
+.. csv-table::
+   :header: "#", "Name", "Function"
+   :widths: 5 25 70
+
+   "MCP", ":hwlabel:`MCP`", "Switch between layers"
+   "Left Strip", "", "Change the tempo of the selected deck (will return to normal on release)"
+   "Right Strip", "", "Change the tempo volume of the selected deck"
```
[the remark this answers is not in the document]
SamWhited: Oops, good catch; fixed!

Swiftb0y (commented):
I have no obvious complaints. Some of the mappings don't seem entirely intuitive to me as they just seem to follow the labels on the controller which is not really made for DJing. In a follow up, we should consider whether its possible to come up with an alternative mapping that better fits the DJing usecase.

SamWhited:
> Some of the mappings don't seem entirely intuitive to me as they just seem to follow the labels on the controller which is not really made for DJing.

Let's chat about this on the forum post at some point; there's definitely plenty of room for improvement!

Swiftb0y:
I'm not active on the forum, but feel free to iterate there with other users. I'll be here and on Zulip to for any code review needs.

Swiftb0y:
@ronso0 anything left here, or is this a LGTY?

Swiftb0y:
In the interest of not letting this go stale, I'll go ahead and merge now.

PR #743, opened 2025-03-13
cleanup and update controller mappings info

ronso0:


source/chapters/controlling_mixxx.rst
```diff
@@ -132,20 +132,10 @@ Supported controllers
...
 MIDI/HID mapping file to tell Mixxx how to understand it. Mixxx comes bundled
-with a number of mappings for various devices. There are two levels of
-controller mappings:
-
-* **Mixxx Certified Mappings**: These mappings are verified by the Mixxx
-  Development Team.
-* **Community Supported Mappings**: These mappings are provided and have been
-  verified as working by the Mixxx community, but the Mixxx Development Team is
-  unable to verify their quality because we don't have the devices ourselves.
-  They might have bugs or rough edges.
```
Swiftb0y, line 143: I think we should mention that by default all mappings are essentially community supported mappings (or should be treated as such). Otherwise ppl may expect the "certified mappings" level support for all currently included mappings.

ronso0: I'm not sure I understand.
If we don't make the distinction between certified and community-supported anymore (and the wiki page with that outdated table wasn't linked anyway), where would the 'certified' expectation come from?

I mean, even though built-in mappings vary greatly in terms of (code) quality, we don't have a quality rating or such for mappings, so if someone discovers issues we'll deal with it individually for each issue, don't we?

ronso0: Okay, I've reverted that.

ronso0: Now this is just a cleanup and a small, helpful addition.
Ready to merge IMHO

Swiftb0y: yup, I agree. still merge conflicts though.

Swiftb0y (approved):
Thank you.

PR #784, opened 2025-08-18
Updated Supported Timecode media table

JoergAtGithub:
-Added information of the production years, that should make clear for most users which version they have.
-Added physical media sizes
-Added info about some encodings according to the xwax page
-Added info that Traktork Mk2 support comes with 2.7 (as the PR with this feature is already merged to main)


ywwg (approved):
LGTM!

PR #793, opened 2025-09-15
feat: add dev container and watch mode

acolombier:
Improve the development experience on the manual


Swiftb0y (commented):
LGTM, though I have no opinion on the devcontainer stuff.

ronso0:
What is watch mode?

daschuer:
Can you explain the issue this is solved here? 
We have the same issue like in the mixxx repository, that the user is urged to use devcintainer even if it is not required or beneficial.

PR #794, opened 2025-09-17
document BPM const filter

ronso0:
docu for https://github.com/mixxxdj/mixxx/pull/14642

### DIFF
--- a/source/chapters/library.rst
+++ b/source/chapters/library.rst
@@ -471,6 +471,14 @@ Mixxx supports the following filters:
 
        ~bpm:100
 
+  * Constant BPM search finds tracks that were analyzed with "Assume constant tempo".
+    Combine with "not"/negative prefix``-`` to find all tracks with variable tempo.
+
+    ::
+
+       bpm:const
+       bpm:constant
+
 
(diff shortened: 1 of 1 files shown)


source/chapters/library.rst
```diff
@@ -471,6 +471,13 @@ Mixxx supports the following filters:
 
        ~bpm:100
 
+  * Constant BPM search finds tracks that were analyzed with "Assume constant tempo".
+    With "not"/negative prefix``-`` it finds all tracks with variable tempo.
+
+    ::
+
+       bpm:const|constant
+
```
Swiftb0y, line 481: NIT:
```suggestion
  * Constant BPM search finds tracks that were analyzed with "Assume constant tempo".
    Combine with "not"/negative prefix``-`` to find all tracks with variable tempo.

    ::

       bpm:const
       bpm:constant
```

ronso0: oki, done.

Swiftb0y: I also had reformulated the second sentence. Sorry, should've mentioned it since the diff doesn't highlight it.

PR #823, opened 2025-12-26
pre-commit: ignore unreferenced Hyperlink targets

absorb-it:
Ignore wrong reports about unreferenced Hyperlink targets - see also https://github.com/rstcheck/rstcheck/issues/77

Sometimes it makes sense to have targets for future referencing included, but even used hyperlink targets are reported as errors. Without this change it is impossible to write or update any documentation - please change this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,5 +53,8 @@ repos:
   hooks:
   - id: rstcheck
     name: "Check RST syntax"
+    args: [
+      --ignore-messages=(Hyperlink target.*not referenced)\.$,
+    ]
     additional_dependencies:
     - sphinx


ronso0:
> Without this change it is impossible to write or update any documentation

When I run into this I check if the current file is affected and if i can fix it.
If not, I use SKIP (one time ignore) and run `SKIP=rstcheck [git command]`. It can also be added to a global ignore list, but I don't rememer the command.

ronso0:
IIUC all controller pages have a reference at the top so the page is listed in https://manual.mixxx.org/2.5/en/hardware/manuals.html ??
Maybe it's just that what's broken?


> Sometimes it makes sense to have targets for future referencing included

We shouldn't add unused references. YAGNI : )
Add it when it's needed.

absorb-it:
> > Without this change it is impossible to write or update any documentation
> 
> When I run into this I check if the current file is affected and if i can fix it. If not, I use SKIP (one time ignore) and run `SKIP=rstcheck [git command]`. It can also be added to a global ignore list, but I don't rememer the command.

seems like overkill to skip complete rstcheck, just to be able to upload and than collide with your pre-commit check again.

PR #835, opened 2026-01-20
Credits drain

daschuer:
This enables client sice caching and stops building manual for old versions. 
This shall safe 50 % of bandwidth and request and shall save use 15 € / month.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -49,3 +49,20 @@
 
     # Stop page from loading if a reflected XSS attack is detected
     X-XSS-Protection = "1; mode=block"
+
+[[headers]]
+  for = "/**/*.woff2"
+  [headers.values]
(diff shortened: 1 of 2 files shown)


JoergAtGithub (changes requested):
We have enough money to keep our services running until we selected a new webhoster. There is no immediate reason to remove the pages which our user access often.

ronso0:
How to proceed here @mixxxdj/developers ?
Is this still required with the resolved Netlify bandwidth issue?

daschuer:
The urgent demand is gone. We can however review this under the fair use of resources aspect. 

Question: 
* What is the oldest manual we need to host and for whom.
* How to solve the canonical link issue, that Google always referes the "current" release manual. 
* Do we want to enable client side caching?

PR #839, opened 2026-01-26
Fix grammar and typos

jerobado:


acolombier (approved):
Good catch! Thanks for fixing it.

PR #712, opened 2025-01-07
Markdown support

JoergAtGithub:
The example-setups page in this PR is generated from MarkDown syntax instead of RST syntax: https://deploy-preview-712--mixxx-manual.netlify.app/chapters/example_setups

PR #855, opened 2026-03-21
Migrate mixxx wiki

acolombier:
Depends on #712 

This is a work in progress effort to move the remaining page from the Mixxx Wiki into the manual.

## Why?

The Mixxx Wiki is a read-only, developer-oriented, legacy, documentation center. User won't be able to neither discover these documentations easily or edit incomplete manuals. They also clog the wiki, making it hard to navigate or search development resources. 

## How?

Thanks to the work done to support Markdown, we can migrate those as-is in the manual. Since some of these are very old, we would add a `Warning` callout on top to raise to the user that the information in here may be incorrect or outdated, and call for action into _TBD_`
In the Mixxx wiki, we will override all these page with a `Moved to <Link>` for sometimes (_TBD_) to ensure that eny existing link would be updated.

## TODO

- [ ] Some mapping duplicate still need to be diffed to find whether the manual or the wiki contains the most info 
- [ ] There is **a lot** of images, some look safe to use (mapping, schema, diagrams), some look potentially stolen from vendor's website. Do we want to review them one by one? 
- [ ] Some existing manual document about mappings not included in the codebase (anymore?) tho mapping can still be found (e.g directly uploaded into the wiki as `.zip` or `.xml`). Remove? Or do we persist the mapping in the codebase with a big fat warning?


ronso0:
Thank you!

Just stumbeld across one we don't need

### Mappings that don't need to be migrated
- [ ] [DJ TechTools MIDI Fighter Twister](https://github.com/mixxxdj/manual/pull/855/changes#diff-0fea34b1cde3dfccd83f7d4e619ec1c49b0e97b45c738bbd9eca8154d73aeee7) -> we already have a [better version](https://manual.mixxx.org/2.5/en/hardware/controllers/dj_techtools_midi_fighter_twister)
- [ ] ...

We can move this to the PR description, and maintainers can just add similar cases.
If in doubt, just add a comment?

acolombier:
Yep, thanks @ronso0 ! There is a few here that actually aren't outdated, and which are fresher/better in the current Manual. I'll try and spin up a table to help with the review/decision!

acolombier:
I have rebased with the recently migrated codebase using Markdown and run some AI to undergo the integration of changes from the Wiki. Not really sure how we could go about this: many documentations still looks quite rough around the edge and seem to highlight a potentially low value.

For reference, here is the prompt I have used, for each files that needed merging (meaning it was already on the manual but also had some information on the wiki)

> You are going to help resolving merge conflict in the hardware manual file <path/to/md>. These conflict have arised from merging an old, often outdated source, which is a Github wiki (the 'Incoming' change from migrate_mixxx_wiki), and the currently live version (the current, or HEAD), often in a cleaner state in terms of presentation. When the wiki information is already captured in the currently live version, simply accept the incoming changes (that is, keep the live version in place). When the wiki version has additional information, incorporate those changes in. Note that the wiki is many years behind, so when the current version suggest contradicting informations, it should be considered more up to date that the wiki version. Do not rewrite any existing information, and keep it as unchanged as possible. You may use connector words and any minimal adjustement to make the wiki content fit in the live version when necessary. Note that https://www.mixxx.org/forums/ doesn't exist anymore so the wiki will contains dead links to it but they should be removed or ignored. https://mixxx.discourse.group/ is the currently live forum and links there are valid.


Happy to run new prompts (e.g "rework the presentation to align better with other manual" if we thing it needs some more changes). Perhaps we could shortlist those we are happy to keep as is? Or we can go as @ronso0 suggested and rather list the one to exclude.

ronso0:
Thank you for the update!
I just started to skim through the changes and noticed for example that there is no mapping (anymore) for the Behringer CMD DV-1
[new manual page](https://github.com/mixxxdj/manual/pull/855/changes#diff-766a2ce981d074de820d8be2e8fbdcedee979e05e5f3c231bd34a48af1cdfe6d)
[mapping dir](https://github.com/mixxxdj/mixxx/tree/2.6/res/controllers)

Same for the Behringer MM1 "advanced mapping" and Behringer CMD Studio 2a if I'm not mistaking.
I assume they were either in the repo at some point, or refer to community mapping that were never merged or only existed in the forums.

What's the smartest way to deal with these? AI or manual check?

acolombier:
Probably best to do a manual check! I'll see what I can do there.

ronso0:
Cool.
Is it intentional that we have multiple pages for one controller?
https://deploy-preview-855--mixxx-manual.netlify.app/hardware/manuals

There are 3 pages for the Behringer MM-1 where only the first (the original?) seems to make sense / is looking good. For 2 and 3 the images are not shown.

Just stumbled across this because there are maaany images in the diff for the MM-1

JoergAtGithub (changes requested):
I'm sorry to have to intervene here, but we can't merge the wiki content in an automatic batch process, because we don't have signed CLAs from many of the former contributors.
Before a page can be merged into the manual, it's licensing situation must be clarified.

acolombier:
At this point, I am now looking into using the content of the wiki as AI context to help fixing the manual and adding potentially missing info, rather than merging as it. This is because the "editorial" aspects if those wiki pages is very inconsistent.
If I am honest, there isn't much yet, except for very targeted and specific version (e.g a link to the forum with an alternative mapping). My plan is to iron every out, and finish the migration to Markdown (which will be cherry picked to another PR). We can then decide, what is worth keeping, what is not, what is subject to license, etc.

acolombier:
Slightly reviving this - I am not sur the CLA argument really hold. We also have mapping written by member of the community ([e.g](https://github.com/mixxxdj/manual/pull/855/changes#diff-1037f7e8508d041f776d742a11b281aa08ce359f6b3eb91346e2bb4c5677ff89)) who have signed the CLA, so even if we decide to hold on some of this mapping, it would be good to still decide whether we want to migrate some of these mapping so we can start cleaning up the `mixxx` wiki. 

We could also decide to leave this effort here, but I would want to consider cleaning the wiki anyway (e.g archive the wiki repo into something like `mixxxx/wiki-mapping-archive`)

JoergAtGithub:
The Denon Prime 4 mapping is not contained in Mixxx. Therefore it should not appear in the manual.

A Wiki article (or section of it) where all editors signed the CLA should be fine to be migrated to the manual.

A wiki cleanup is neccessry, and I consider the current mixxxdj/wiki already as archive, as it's read only for most people. The biggest problem is https://github.com/mixxxdj/mixxx/wiki/Hardware-Compatibility which is totally outdated, but is our most often clicked URL.

acolombier:
Closing for now then. I will create a proposal to suggest how to properly clean up the wiki later.

PR #858, opened 2026-03-28
Backport markdown support to 2.6

JoergAtGithub:
If we introduce the new syntax in `main` first we get a maintenance problem, because updates of the `2.6` manual could be in RST syntax, while the same page in `main` might be already converted to MD syntax. The result would be lots aof merge conflicts.


daschuer (approved):
LGTM and sound resonable.

PR #859, opened 2026-03-28
Enable dependabot for precommit

JoergAtGithub:
- Enable dependabot for precommit
- Set dedeicated target branch
Same settings as in mixxxdj/mixxx


acolombier (approved):
We should also consider doing this for PIP, though we currently don't pin dependencies, so likely for a follow up!

PR #860, opened 2026-03-28
Only run linkcheck for changed source files in PRs

JoergAtGithub:
This still runs the full link-check over all source files on push commits, but in PRs it only checks th links in files source files changed compared to the base commit of the PR.
Means if we want to see the full list of broken links, we can always look at the resul of branches like 2.5, 2.6 or main, while the CI result of PR buils will be green unless a link in the modified fils is broken.

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -11,16 +11,31 @@ jobs:
     steps:
     - name: Clone repository
       uses: actions/checkout@v6
+      with:
+        fetch-depth: ${{ github.event_name == 'pull_request' && 0 || 1 }}
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
         python-version: '3.13'
+    - name: Fetch base branch for diff
+      if: github.event_name == 'pull_request'
+      run: git fetch origin ${{ github.event.pull_request.base.sha }}
+    - name: Get changed source files
+      id: changed
+      if: github.event_name == 'pull_request'
+      env:
+        GITHUB_EVENT_NAME: ${{ github.event_name }}
+        PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        PUSH_BEFORE_SHA: ${{ github.event.before }}
+      run: python tools/list_changed_source_files.py
     - name: Install Python dependencies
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
I tested this with injected failures in this PR. Finally it worked as intended and I reverted the injected failures.

.github/workflows/sphinx.yml
```diff
@@ -11,16 +11,31 @@ jobs:
...
+      run: python tools/list_changed_source_files.py
     - name: Install Python dependencies
+      if: github.event_name == 'push' || steps.changed.outputs.files != ''
       run: |
         python -m pip install --upgrade pip
         python -m pip install -r requirements.txt
     - name: Check for broken links
-      run: sphinx-build -q --color -b linkcheck source build
+      if: github.event_name == 'push' || steps.changed.outputs.files != ''
+      run: sphinx-build -q --color -b linkcheck source build ${{ steps.changed.outputs.files }}
```
acolombier, line 38:
```suggestion
    - name: Fetch base branch for diff
      if: github.event_name == 'pull_request'
      run: git fetch origin ${{ github.event.pull_request.base.sha }}
    - name: Install Python dependencies
      if: github.event_name == 'push' || steps.changed.outputs.files != ''
      run: |
       python -m pip install --upgrade pip
       python -m pip install -r requirements.txt
    - name: Check for broken links
      env:
        BEFORE: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
      run: |
        changed_files=`git diff $BEFORE --name-only --diff-filter=d HEAD -- source/ | grep -E '\.(md|rst)$' | xargs`
        if [ -n "$changed_files" ]; then
          sphinx-build -q --color -b linkcheck source build $changed_files
        fi
```

acolombier (commented):
Thanks for addressing this. I have suggested a bash-only approach, that could allow removing the Python script and allow to easily run it locally when debugging

JoergAtGithub:
bash is Unix only. Not a good option for a platform Independent project.

acolombier:
Does git not come with Bash anymore on Windows?

PR #861, opened 2026-03-28
Add canonical links

JoergAtGithub:
Add canonical link to the manaul of the latest released Mixxx version to each HTML page

Closes: #838

### DIFF
--- a/build_html.sh
+++ b/build_html.sh
@@ -4,7 +4,6 @@
 # and want to use "en" as a fallback.
 cd source || exit 1
 LANGUAGES="$(python -c 'import conf; print(" ".join(sorted((lang for lang in conf.supported_languages.keys() if lang != "en"), reverse=True)))')"
-SMV_LATEST="$(python -c 'import conf; print(conf.smv_latest_version)')"
 cd .. || exit 1
 LANGUAGES="$LANGUAGES en"
 NUM_LANGUAGES="$(printf '%s' "$LANGUAGES" | wc -w)"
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
Seems to work, each Netlify built HTML page contains now a line like:
`<link rel="canonical" href="[https://manual.mixxx.org/2.5/en/index.html](view-source:https://manual.mixxx.org/2.5/en/index.html)" />`

daschuer:
I am new to this, but reading the linked wiki page it looks like that ref="canonical" means: 
If you find two web pages with identical content, prefer the one mentioned as "canonical" 
In case of 2.4 and 2.5 links, the content might be identical, but not necessarily. 

After a version bump, we have to move the canonical link, which is probably not flawless, because we originally told Google that 2.5 is canonical and the index is still populated and ranked for the old. I have no idea how google handles that. 

We already have such an identical duplicate situation: 
https://manual.mixxx.org/2.5/chapters/introduction
points to https://manual.mixxx.org/2.5/**de**/chapters/introduction

How about this: 
Add an alias that replaces the the 2.5 part by a "current" or "stable" or just removes it. 
Make this the ref="canonical" and add "noindex" to the versioned URLs. 
This way we have actually URLs pointing to identical content. 
https://manual.mixxx.org/chapters/introduction
https://manual.mixxx.org/**2.5/de**/chapters/introduction

This way we have only "stable" links in the Google index which will automagically redirect to the latest stable after a version bump.

JoergAtGithub:
> How about this:
> Add an alias that replaces the the 2.5 part by a "current" or "stable" or just removes it.
> Make this the ref="canonical" and add "noindex" to the versioned URLs.

Good idea, it seems that we have already such an alias. The following links work for me:
https://manual.mixxx.org/latest/hardware/manuals
https://manual.mixxx.org/latest/fr/hardware/manuals

daschuer:
Cool, so this should be our canonical link? 
How do we tell Google to only index this?

JoergAtGithub:
> How do we tell Google to only index this?

We don't need to. With this PR, each HTML page contains a statement like
 `<link rel="canonical" href="[https://manual.mixxx.org/latest/en/hardware/manuals.html](view-source:https://manual.mixxx.org/latest/en/hardware/manuals.html)" />`
in the HEAD.
Next time the Google crawler reads this page, it will evaluate the canonical link.

JoergAtGithub:
Can we merge this?

PR #865, opened 2026-04-06
(fix) Library: remove 'renamed' from rescan description

ronso0:
as pointed out in https://github.com/mixxxdj/mixxx/pull/14507#issuecomment-4194535538

Currently the scanner only [compares filename and duration](https://github.com/mixxxdj/mixxx/blob/01a689c0b4d99cf0db6079d134cfed44aef36c5b/src/library/dao/trackdao.cpp#L1921) of missing and new tracks, so this is misleading.


daschuer (approved):
LGTM, Thank you.

PR #868, opened 2026-04-10
Update Reloop Terminal Mix 2/4 for JS components

git-developer:
See https://github.com/mixxxdj/mixxx/pull/16316

The documentation of SHIFT + LOOP IN / LOOP OUT ("drags the loop point of an active loop") was removed because that mapping was not implemented.


acolombier:
We have now migrated the hardware documentation to Markdown, as we are looking to consolidate it in the main repo in the future. Some conflicts have arise from this change. Could you please update your changes to use Markdown instead?

git-developer:
Done

PR #873, opened 2026-04-14
Search: adjust result ranking, push hardware manual hits down

ronso0:
This aims to improve the ranking when searching the manual.
Currently, when you search for `gain`, a lot of hardware manuals are at the top -- very likely not what the user wants to see.

New priorities are
0 chapters/
1 chapters/appendix/
2 hardware/

### before:
https://manual.mixxx.org/2.5/en/search.html?q=gain&check_keywords=yes&area=default

### this PR:
https://deploy-preview-873--mixxx-manual.netlify.app/search.html?q=gain&check_keywords=yes&area=default

___
disclaimer: created the inject/override with an agent, so bare with me when it comes to js nitpicking ; )

### DIFF
--- a/source/_static/js/search-scorer.js
+++ b/source/_static/js/search-scorer.js
@@ -1,0 +1,48 @@
+/**
+ * Adjust Sphinx Search Scoring
+ * This script overrides the default Sphinx scoring to de-prioritize hardware manuals
+ * and control documentation.
+ *
+ * https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_search_scorer
+ *
+ * Priorities:
+ * 0 chapters/
+ * 1 chapters/appendix
+ * 2 hardware manuals
+ */
+var Scorer = {
+    score: result => {
+        const [docName, title, anchor, descr, score, filename] = result;
+        let adjustedScore = score;
+        if (docName.includes('hardware/')) {
+            adjustedScore -= 100;
+        } else if (docName.includes('chapters/appendix/')) {
+            adjustedScore += 50;
+        } else { // docName.includes('chapters/')
+            adjustedScore += 100;
+        }
+        return adjustedScore;
+    },
+
+    // Standard Sphinx weights
+    // query matches the full name of an object
+    objNameMatch: 11,
+    // or matches in the last dotted part of the object name
+    objPartialMatch: 6,
+    // Additive scores depending on the priority of the object
+    objPrio: {
+      0: 15, // used to be importantResults
+      1: 5, // used to be objectResults
+      2: -5, // used to be unimportantResults
+    },
+    //  Used when the priority is not in the mapping.
+    objPrioDefault: 0,
+
+    // query found in title
+    title: 15,
+    partialTitle: 7,
+
+    // query found in terms
+    term: 5,
+    partialTerm: 2,
+};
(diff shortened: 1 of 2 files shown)


source/_static/search_adjust.js
```diff
@@ -0,0 +1,49 @@
+
+
+
+
+
```
daschuer: Stray new lines.

source/_static/search_adjust.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Mixxx Manual Search Optimizer -  Adjust Sphinx Search Scoring
+ * This script overrides the default Sphinx scoring to de-prioritize hardware manuals.
+ * Priorities:
+ * 0 chapters/
+ * 1 chapters/appendix (because of noise from mixxx_controls.rst)
+ * 2 hardware manuals
+ */
```
daschuer: Please add the manual link: 
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_search_scorer

source/_static/search_adjust.js
```diff
@@ -0,0 +1,49 @@
...
+var Scorer = {
+    score: function(result) {
+        const docname = result[0];
+        let currentScore = result[4];
```
daschuer: This looks nicer, taken form the manual, without magic numbers: 
```suggestion
    score: result => {
      const [docName, title, anchor, descr, score, filename] = result
```

source/_static/search_adjust.js
```diff
@@ -0,0 +1,49 @@
...
+
+        // TIERED RANKING
```
daschuer: Remove?

source/_static/search_adjust.js
```diff
@@ -0,0 +1,49 @@
...
+
+        // Priority 1: Appendix
+        else if (docname.includes('chapters/appendix/')) {
+            currentScore += 50;
+        }
+
+        // Priority 2: Hardware (The "Noise")
+        else if (docname.includes('hardware/')) {
+            currentScore -= 100;
+        }
```
daschuer: If you revert the order, you can get rid of one `docname.includes('chapters/appendix/')`

source/_static/search_adjust.js
```diff
@@ -0,0 +1,49 @@
...
+
+        return currentScore;
+    },
+
+    // Standard Sphinx weights
```
daschuer:
```suggestion
    // Standard Sphinx weights found in https://github.com/sphinx-doc/sphinx/blob/cc7c6f435ad37bb12264f8118c8461b230e6830c/sphinx/themes/basic/static/searchtools.js#L22
```

Maybe we want to copy them literally including the comments?

daschuer (commented):
While it works, it is probably better to follow more closely the example form the manual.

daschuer:
https://deploy-preview-873--mixxx-manual.netlify.app/search.html?q=Hercules&check_keywords=yes&area=default#

This a special case because of the deprecated controls. 
But do we want to lower the changelog matches?

source/_static/search_adjust.js
ronso0: ..yes, better.
(but requires an extra `let adjustedScore`)

source/_static/search_adjust.js
ronso0: I'll use the default text from https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_search_scorer

daschuer (approved):
LGTM, Thank you.

PR #878, opened 2026-04-21
Fix most dead links in controller mappings

Krafting:
Some controller mapping have dead link, I replaced them with their current counterparts. 

For some, the official website is not available anymore, I replaced them with the internet archive version instead, as it was done for other mappings

I hope this helps!


daschuer:
Some parts are also in this hidden PR. I will update it accordingly.

daschuer (approved):
LGTM, Thank you.

PR #883, opened 2026-05-10
Xone K3 docs

Reasonable-Solutions:
allen & heath released the xone k3 last year. It's basically the same as the k2 but no audio interface. After updating the firmware it works in exactly the same way as the other k controllers.

### DIFF
--- a/source/hardware/controllers/allen_heath_xone_k2_k1.rst
+++ b/source/hardware/controllers/allen_heath_xone_k2_k1.rst
@@ -1,14 +1,16 @@
 .. _allen-heath-xone-k2:
+.. _allen-heath-xone-k3:
 
-Allen & Heath Xone:K2
-=====================
+Allen & Heath Xone:K2/K1/K3
+===========================
 
 The Allen & Heath Xone:K2 is a flexible controller that is mapped to
 control Mixxx’s decks, effects, or both. It has an integrated 4 channel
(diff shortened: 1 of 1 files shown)


ronso0:
Thank you for the update!

Do you mind to update the MIDI mapping description as well?
https://github.com/mixxxdj/mixxx/blob/4cb57aa65499b21deb1d2be95a6a487dbaa18e5b/res/controllers/Allen%20and%20Heath%20Xone%20K2.midi.xml#L7

And the paperwork...
As a first-time contributor we need you to sign the [Mixxx Contributor Agreement](https://docs.google.com/forms/d/e/1FAIpQLScC9QG327sjLL0eWftmfGUasxFFLxg6LCXJ2xHDYRzFIRqyiw/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ) and comment here when you have done so. It gives us permission to distribute your contribution under the GPL v2 or later license and the Apple Mac App Store. It is also helpful for us to have contact information for contributors in case we may need it in the future.

Reasonable-Solutions:
contributor agreement handled

ronso0:
LGTM, only thing my brain stumbles across is the order K2 / K1 / K3
Do you mind changing that to 1-2-3?

ronso0 (approved):
Thank you, LGTM

PR #867, opened 2026-04-09
Migrating hardware manual to Markdown (first batch)

acolombier:
This migration was automated with a Python script, largely written by an AI and manually tweaked. 

- AFAIS, there is no (or very limited) translation in the manual, but this remains something to discuss.
- Some build error (notably `Non-consecutive header level increase`) are existing issue in RST. We can decide to ignore those or fix in the migration.

---

## Test

Migrated pages were reviewed side-by-side with the live RST version on the [website manual](https://manual.mixxx.org/2.7/en/) by eyes, so mistake might have sneaked in. When there was too many visual discrepancy, the manual was discarded from this batch, for later review. 

## Note
- This is the first batch of manual, where no visual regression were found. Note that table column size was **not** considered as regression and thus the new Markdown manual is not pixel perfect to the previous RST version, as it would look like the auto-sizing algorithm between RST and MD is not the same. 
  When the column layout was explicitly, this explicit declaration was kept in Markdown
- Many visual bugs were found in the previous RST version (label not rendering, unclosed formatting, broken links or lists). Most of those were corrected automatically, since Markdown is more permissive than RST, but some are still a problem.

## Plan and future steps

The idea is to carry on with the Markdown migration, before those can all moved (at once or by batches) to the main repo, and pulled automatically (either with a repo automation or at build time). 
Once moved into the main repo, we will be able to better help contributor to keep the doc up to date and use AI to help reviewing some poor or incomplete documentation, using the actual mapping definition. 

Preview available [here](https://deploy-preview-867--mixxx-manual.netlify.app/hardware/manuals)

PR #708, opened 2024-12-29
pre-commit: Ignore time stamps in po files

daschuer:
I revert normally manually time stamp only changes, to avoid cluttering a commit wit 4100 changed files. This is tedious and error prone. This PR automates it.

PR #891, opened 2026-06-20
chore: migrate hardware manual to Markdown (part 2)

acolombier:
Follow up on #867, extracted from #855


JoergAtGithub:
Please fix the build check fails!
Are there any significiant sections taken from the Wiki?

acolombier:
> Please fix the build check fails! 

Just forced pushed a fix, looks like some syntax messed sneaked in.
PDF is still borken, I will investigate later.
 
> Are there any significiant sections taken from the Wiki?

None! This aims to be a "content perfect" migration, so we can simplify the review of #855 later.
For disclosure, I have used an agentic approach and used [Playwright MCP](https://github.com/microsoft/playwright-mcp) for the success loop. It would compare the rendering of the local server (using sphinx-autobuild) and the 2.6 server.

Note that it isn't perfect. I could see some mild discrepancy so it will still need a human review to catch those regressions that we consider problematic.

source/hardware/controllers/american_audio_vms2.md
```diff
@@ -10,18 +9,18 @@ can also be used as a stand-alone mixer with analog sources without a
 computer by setting the USB/Analog switches on the front of the device
 to “Analog”.
 
--  `Manufacturer’s product page <http://www.americandj.eu/en/vms2.html>`__, provides manual and drivers for download.
--  `Forum thread <https://mixxx.discourse.group/t/american-audio-vms2/12000>`__, for discussion of mapping options.
--  `Pull Request <https://github.com/mixxxdj/mixxx/pull/876>`__, which this wiki page describes.
+-  [Manufacturer’s product page](http://www.americandj.eu/en/vms2.html), provides manual and drivers for download.
+-  [Forum thread](https://mixxx.discourse.group/t/american-audio-vms2/12000), for discussion of mapping options.
+-  [Pull Request](https://github.com/mixxxdj/mixxx/pull/876), which this wiki page describes.
```
ronso0:
```suggestion
```

source/hardware/controllers/behringer_bcr2000.md
```diff
@@ -0,0 +1,234 @@
...
+   :align: center
+   :width: 90%
+   :figwidth: 100%
+   :alt: Behringer BCR2000 (schematic view)
+   :figclass: pretty-figures
+
+   Behringer BCR2000 (schematic view)
+```
+
+- [Manufacturer's product page](https://www.behringer.com/behringer/product?modelCode=P0245)
```
ronso0:
```suggestion
- [Manufacturer's product page](https://www.behringer.com/en/products/0808-AAA)
```

source/hardware/controllers/behringer_ddm4000.md
```diff
@@ -0,0 +1,211 @@
...
+used either for audio or as MIDI controller:
+
+* Channel 1-4
+* Microphone Channel
+* Crossfader
+* Sampler
+
+The mixer contains no digital interfaces for audio or microphones.
+
+* [Manufacturer's product page](https://www.behringer.com/behringer/product?modelCode=P0167)
```
ronso0:
```suggestion
* [Manufacturer's product page](https://www.behringer.com/en/products/0701-AAK)
```

ronso0 (commented):
just some fixes for broken links I noticed

acolombier:
Thanks @ronso0 !

I spend a good three hours to do a side by side comparison and update small details to get a near pixel perfect migration (+ fixing rendering issues in the live version).

I had to do two side chnages which are:
- Integrate a temporary fix for https://github.com/return42/linuxdoc/issues/44
- Create a hook to circumvent the [MyST footnote's behaviour](https://myst-parser.readthedocs.io/en/latest/syntax/typography.html#footnotes) (MyST always render footnote at the bottom, where as previous RST was rendered when defined)

Out of the thirty manual updates, here are my notes for future content improvement

> hercules_dj_console_4_mx: Need some care for better rendering (likely automat-able with agentic workflow)
hercules_dj_console_rmx: Fixed random spaces in words
hercules_dj_console_rmx: Need some care for better rendering, especially hwlabel (likely automat-able with agentic workflow)
hercules_p32_dj: fixed image not correctly defined
numark_mixtrack: Need some care for better rendering (likely automat-able with agentic workflow)
numark_mixtrack_platinum: not exactly the same rendering as the original rendering, fixed some issues with bottom lists

Note that I found some content that had been changed by the AI beside the strict rules around respecting a perfect content migration. I have removed those. 

Something remains wrong with the PDF build but not sure what. I need to investigate further. 
**Edit**: seem to be failing on `2.6` already?

ronso0:
> I spend a good three hours to do a side by side comparison and update small details to get a near pixel perfect migration (+ fixing rendering issues in the live version).

Thank you!

So, in terms of review, we'd do just a quick side-by-side check?

And the wiki pages are postponed until we figure the license situation? (as @JoergAtGithub proposed)

acolombier:
> So, in terms of review, we'd do just a quick side-by-side check?

Yep! I have done it already, but a quick check would be good to confirm I didn't miss anything!

> And the wiki pages are postponed until we figure the license situation? (as @JoergAtGithub proposed)

Absolutely, once I will have rebased the wiki migration PR, we should see much clearer as it will reduce the diff set quite a bit and we then decide if there is info worth keeping in (e.g with a AI-enabled rewrite) or simply noise/redundant and drop it, thus consider the Wiki entry as not adding any value (disclaimer, I think this is the case for +90% of it :D )

acolombier:
Ok, everything is fixed. Had to make some fixes to l10n and latex rendering engine since the Markdown plugin (MyST) is more strict about certain errors (e.g not tolerating invalid roles).
We could probably disable those, but I thought this was a great opportunity to clean it up.
Also, only 2 dead link (which aren't actually dead, but blocked because coming from GH CIDR), which is the smallest I have ever seen on the repo :laughing:

JoergAtGithub:
Does it work to edit translated strings here? Wouldn't the next tranifex update overwrite them with the old string again?

acolombier:
I think it would, I believe we would need to import those fix back into Transifex, before further edits are made! 
At the moment, since there were breaking the build, I was left with no choice but fixing it, but we should do the import as fast as possible so no community edit occurs.
Not sure how to do that?

ronso0:
> I think it would, I believe we would need to import those fix back into Transifex, before further edits are made!
At the moment, since there were breaking the build, I was left with no choice but fixing it, but we should do the import as fast as possible so no community edit occurs.
Not sure how to do that?

It is the wrong appraoch to change the .po files manually. It should suffice to change the sources only (.rst files), then update/regerate the tr templates and push to Transifex.
I think the process is described here https://github.com/mixxxdj/manual#maintaining-translations

acolombier:
The problem is there was strings that where translated in Transiflex already, which should remain in English. 

Example:
- ``:menuselection:`Preferences ...` `` -> ``:menuselectie:`Voorkeuren...` `` (see the role `:menuselection:` has been translated)
- `|ic_lib_computer| This mode works... ` -> `|ic_lib_ordinateur| Ce mode fonctionne...` (see the icon role has been translated)
- 
Happy to revert [the commit](https://github.com/mixxxdj/manual/pull/891/changes/2fdf6373a780ae581d05652e4a84e0800454965f#diff-d41bd72fbd19a26b722e2cf87d9c9bc9c672c0dc11f8821094a9faf3e3f1f476) otherwise but this will fail the build. I can perhaps try to filter the error with `suppress_warnings` or similar tho but note that translation breaks the page rendering currently anyway.

Happy to also report those fixes to Transiflex but I don't think I have access.

daschuer:
> Happy to also report those fixes to Transiflex but I don't think I have access.

Anyone has access. I just went ahead and fixed it there for all Versions. 
For updating it to 2.6, we need to merge https://github.com/mixxxdj/manual/pull/708 first.

acolombier:
Good to know! I am really not familiar with the setup regarding i18n

acolombier:
I have now rebased this branch, resoled conflicts from recent changes and remove the translations changes.
@mixxxdj/developers It'd be good if we could merge this ASAP as it will catch conflicts every time a new manual change is merged.

JoergAtGithub (changes requested):
This looks nearly perfect, but I found 1 regression and 2 small readability issues:
The links on top of the DDJ-400 page are no longer correct (Label and link target changed):

Image

Image

--
Linebreaks at the top of Korg Kaoss page got lost:

Image
--
Linebreaks (distance between paragraphs) are remove in https://deploy-preview-891--mixxx-manual.netlify.app/hardware/controllers/numark_mixtrack_platinum#mapping
vs.
https://manual.mixxx.org/2.6/en/hardware/controllers/numark_mixtrack_platinum#mapping
which reduces readability

JoergAtGithub:
@acolombier It would be good, if you could address these minor findings, that we can merge this PR before conflicts occur.

acolombier:
Thanks for the ping, I did missed the notification, will have a look ASAP.

acolombier:
In order to save time, I have pushed a fixup (https://github.com/mixxxdj/manual/commit/bc33a393e52c9b895315f4ce0d4ef8ff24dcd3fe) and force pushed the squash ([null-diff](https://github.com/mixxxdj/manual/compare/bc33a393e52c9b895315f4ce0d4ef8ff24dcd3fe..6272085f1a07ae0b4cecf9e612bc499ddd3414bc))

JoergAtGithub:
The formating of the Numark Mixtrack Platinum is still not correct:
grafik
vs.
grafik

acolombier:
Updated with the same process. (2cfba7d1c)

It is not pixel perfect, but this is as close as I can get without having to spend  more time that I can allocate right now.

JoergAtGithub (approved):
LGTM!Thank you!

ronso0:
Yup, thank you!

PR #922, opened 2026-08-23
Set browser User-Agent for linkcheck to avoid 403 errors

mxmilkiib:
## Summary
- Setting `linkcheck_user_agent` to a real Firefox User-Agent string in `source/conf.py`, fixing the false broken-link warning for `https://yaeltex.com/product/minimixxx/` which returns 403 to the default Python urllib agent but 200 to a browser.
- Addresses the root cause rather than adding the URL to `linkcheck_ignore`, so the link is still actually checked.

#### Test plan
- [x] Ran `sphinx-build -b linkcheck source build` — yaeltex link passes with no warnings
- [x] No new broken links or 403 errors introduced

Generated with [Devin](https://devin.ai)

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -477,7 +477,9 @@ linkcheck_retries = 2
 # Use a real browser User-Agent to avoid 403 errors from sites that block
 # the default Python urllib agent (e.g. Cloudflare-protected sites like
 # yaeltex.com)
-linkcheck_user_agent = "Mozilla/5.0 (X11; Linux x86_64; rv:128.0) Gecko/20100101 Firefox/128.0"
+linkcheck_user_agent = (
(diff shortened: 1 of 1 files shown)


mxmilkiib:
this fixed it locally, but the CI tests are still blocked by Cloudflare because they get detected as automatic, so the link is added to the ignore list

daschuer (approved):
LGTM, Thank you.

daschuer:
The audacity link has moved to: 
https://support.audacityteam.org/au3/troubleshooting/solving-recording-problems/latency-compensation

Would you mind to fix that as well?

daschuer:
The audacity link has been fixed here: 
https://github.com/mixxxdj/manual/pull/937

PR #925, opened 2026-08-23
Sync Lock: clarify it only does tempo macthing in the first sentence

ronso0:
Just a small fix to avoid confusion.

Previously (in https://manual.mixxx.org/2.5/en/chapters/djing_with_mixxx#beatmatching-and-mixing) we explain that there's tempo and phrase matching. At the end of the chapter there's a hint that mentions Quantize.


daschuer (approved):
Makes sense. Thank you.

PR #928, opened 2026-08-29
Move scripting docs from wiki to manual

SamWhited:
This is the start of moving the scripting docs as described in mixxxdj/mixxx#16814

If this is something we actually want to do (I don't believe a consensus was reached), I'd like to start getting this part of it done. I've moved a few files over as an example of where it could live and to allow folks to chime in if they think this should eg. be its own top level chapter, or part of the appendix, etc.

If there are other articles that should also be a part of this, let me know.


acolombier (commented):
Thanks for looking into this! Is there a reason for using RST instead of markdown, which is now supported?

SamWhited:
I had no idea markdown was an option, i just did what every other manual page did.

PR #935, opened 2026-09-11
Remove line that makes the wiki sound up-to-date

SamWhited:
The wiki is likely no longer the cannonical place for controller mapping info since most people can't edit it and many articles have been moved to this manual.
Remove a sentence that makes it sound like the wiki is the more up-to-date resource.


daschuer (approved):
LGTM, thank you.

Issue #650, opened 2024-05-04
Proof read the S4 Mk3 documentation


acolombier:
In order to improve the quality of the documentation for first class Mixxx device, the S4 Mk3 documentation should be refined to ensure a premium support

acolombier:
@ywwg just for tracking

ywwg:
is the request to just read over the existing doc and tidy it up?

acolombier:
That would be great yes. I'm sure there is a few things that could use some better English wording. Also, it isn't impossible that some information is missing or incorrect

ywwg:
I have started working on this, I'll post a PR when I get through it

PR #936, opened 2026-09-12
Edit the first few sections of the S4MK3 manual

SamWhited:
This edits the first few sections for tone and grammar, as well as to add cross-links that may be useful. It also re-wraps lines that were being touched anyways to a column width of 80 per the guidelines.

Finally, it converts several tables to CSV to make them easier to edit in the source file (and because pre-commit was complaining about the formatting of the old tables even though I hadn't touched them).

Notably I did not re-organize or add/remove information from sections, this may still need to be done but the linked issue is a bit vague on the desired outcome.

Updates #650

### DIFF
--- a/source/hardware/controllers/native_instruments_traktor_kontrol_s4_mk3.rst
+++ b/source/hardware/controllers/native_instruments_traktor_kontrol_s4_mk3.rst
@@ -1,27 +1,33 @@
 Native Instruments Traktor Kontrol S4 MK3
 =========================================
 
-The Kontrol S4 MK3 is a 4 deck all-in-one controller with a sturdy build quality and integrated sound card.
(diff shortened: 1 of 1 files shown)


ronso0:
Thank you for taking care. Though, since the hardware manuals were converted to markdown in #867 and #891 this will create many conflicts when merging 2.5 into 2.6.
Can you try to apply the changes to the 2.6 version?
https://github.com/mixxxdj/manual/tree/2.6/source/hardware/controllers

SamWhited:
Oops, I hadn't seen any of this! I'm not sure that I want to go back through it again for yet another custom syntax (the conversion would be easy enough, but all the custom extensions would be a lot of work to go modify again). If I find the time I'll try, but probably not any time soon.

PR #937, opened 2026-09-12
Recommend Tenacity instead of Audacity

SamWhited:
Audacity has removed their latency compensation tutorial article, but Tenacity (the most commonly used fork of Audacity) has a similar article on their help, so recommend that instead to fix the broken link.


daschuer:
We want an instruction to measure the air to air latency. The article has moved here: 
https://support.audacityteam.org/au3/troubleshooting/solving-recording-problems/latency-compensation

Maybe we can put something in our manual that the users can do with Mixxx right away. This is however some more work. 
We have https://mixxx.org/news/2021-05-09-jack-zero-latency/ outdated article. 
With the recent pipewire implementation in Mixxx it gives actually zero additional latency. 

So we may write a new article. Maybe @pri-yan-shu can help with some details.

SamWhited:
Oops, I searched all over their support site and couldn't find a replacement, weird. The Tenacity article is shorter and easier to read with lots less filler, but either seems like it would be fine.

daschuer:
Tenacity is not part of the Ubuntu repositories. It rather points to this: 
https://packages.ubuntu.com/resolute/python3-tenacity
https://tenacity.readthedocs.io/en/latest/

So lets keep audacity for now. Te best solution would be a Mixxx based description.

SamWhited:
This is for OSX, not Ubuntu (but I take your point). Switched back.

daschuer:
Thank you.

